PR #66, opened 2016-07-12
Message Areas/Conference Next / Prev

apamment:
You will probably want to neaten this up or maybe change it entirely, I thought I'd do a pull request so you can see the changes. I don't know why it changes the whitespace, I'm not intentionally doing that.

### DIFF
--- a/core/predefined_mci.js
+++ b/core/predefined_mci.js
@@ -5,6 +5,7 @@
 const Config				= require('./config.js').config;
 const Log					= require('./logger.js').log;
 const getMessageAreaByTag	= require('./message_area.js').getMessageAreaByTag;
+const getMessageConferenceByTag       = require('./message_area.js').getMessageConferenceByTag;
 const clientConnections		= require('./client_connections.js');
 const sysProp				= require('./system_property.js');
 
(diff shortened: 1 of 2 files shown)


NuSkooler:
The looks pretty good (I've yet to try it). The only thing off the top of my head glancing at it is I'd for e.g. use [find()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find) vs looping over the array. 

What do you mean about changing whitespace?

I'll try and merge this tonight!

apamment:
I didn't even know there was a find command. will look it up shortly :)

About the whitespace, when looking at the commit it seems git minuses some blank lines and replaces them with blank lines. Not sure what that's about.

NuSkooler:
Merged to play with. I do think a MenuModule.redraw() would be good so a full pop/push wouldn't need required. I'll prob. play with that. 

If you convert to find/findIndex/whatever please send a new PR and I'll get it pulled in.

apamment:
> Merged to play with. I do think a MenuModule.redraw() would be good so a full pop/push wouldn't need required. I'll prob. play with that. 

Yes, that would be a good addition, and make things easier in the future :smile: 

Haven't looked at the find command yet, will do so now.

PR #67, opened 2016-07-13
Msg area next/prev updated to use findIndex

apamment:
Updated to remove the for and replace with findIndex.


NuSkooler:
Works well - updated a bit in master. I'll try to come up with a clean redraw-only solution at some point.

Issue #44, opened 2016-01-18
Create a installation script


NuSkooler:
It would be nice to have a installation script that users -- especially those completely new to Node.js -- can use to quickly get up and running.

Example:
```
curl -o- https://github.com/NuSkooler/enigma-bbs/misc/install.sh | bash
```
(Note: **Just an example!** The above does not currently exist!)

The script should try to work on most popular *nix environments. Windows will require something else...

NuSkooler:
Probably provide two paths/choices:
  1. With [nvm](https://github.com/creationix/nvm)
  2. Standard Node.js installation

Should **never** install requiring root

tracker1:
For windows, you could setup build/release that could zip up the entire thing, after npm install, with a copy of node.exe in the project directory... 

Alternative might be to use electron as a base, and then could integrate UI features, since most windows users are likely to run this under the context of a specific user.

Additional reference:
[1] https://github.com/stephenwvickers/node-os-service

NuSkooler:
I think for now this should be a *nix task with a separate Windows task. Reason being that the source should be distributed in all cases with the ability to git pull or npm update... which of course is a issue if binaries are distributed in order to not require Visual Studio or the like. 

Perhaps when we're at the 1.0 point (e.g. out of alpha/beta) releases could be provided in more packaged up forms.

NuSkooler:
> [1] https://github.com/stephenwvickers/node-os-service

This is looks pretty cool for Windows. I'm currently running Xibalba via [PM2](https://github.com/Unitech/pm2) on CentOS.

ellisgl:
If you don't need load balancing and your version of Linux runs Systemd, you can just create a Systemd profile for it.

NuSkooler:
> If you don't need load balancing and your version of Linux runs Systemd, you can just create a Systemd profile for it.

Yeah I like having some of the additional pm2 functionality. This kinda stuff does seem like it should be put in the docs somewhere.

davestephens:
I'm gonna have a stab at this, I was thinking about doing it anyway just so I can reinstall if/when my Pi dies...

NuSkooler:
@davestephens FWIW, I'd like to stick with installing Node itself via `nvm` as installing Node.js via package management and the like should be banned from the earth (installs stuff as root/etc.). This would ensure the entire process does not need root/sudo and Node versions can be switched easily.

[...]

PR #87, opened 2016-08-23
First stab at an ENiGMA installer

davestephens:
ENiGMA installer first cut - per my comments on issue #44.


NuSkooler:
Awesome! I can't right at this moment, but I will give this some runs on my systems over the next couple days.

misc/install.sh
```diff
@@ -0,0 +1,108 @@
+#!/usr/bin/env bash
+
+ENIGMA_NODE_VERSION=4.4
```
tracker1: If you just use `4` you'll now get `4.5`, where `4.4` will only give you `4.4.x`

davestephens: 4.4 was intentional based on the current installation instructions - will
wait for Bryan to say confirm testing Enigma with 4.5 before changing
anything...

NuSkooler:
I've tested most versions below 6. That will be the next major upgrade, which will honestly be very nice to move to. Any of the 4.x versions should be fine though, so having it go to the latest LTS is probably the right direction.

Going to toy around with this in a few minutes & try to come up with comments/etc.

misc/install.sh
```diff
@@ -0,0 +1,108 @@
...
+                                                                     ______
+_____________________   _____  ____________________    __________\\_   /
+\\__   ____/\\_ ____   \\ /____/ /   _____ __         \\  /   ______/ // /___jp!
+ //   __|___//   |    \\//   |//   |    \\//  |  |    \\//        \\ /___   /_____
+/____       _____|      __________       ___|__|      ____|     \\   /  _____  \\
+---- \\______\\ -- |______\\ ------ /______/ ---- |______\\ - |______\\ /__/ // ___/
+                                                                       /__   _\\
+       <*> ENiGMA½ // https://github.com/NuSkooler/enigma-bbs <*>        /__/
+
+This script will install Node 4.4 via nvm, download Enigma½, install its dependencies,
```
NuSkooler: Should ref `ENIGMA_NODE_VERSION` here

misc/install.sh
```diff
@@ -0,0 +1,108 @@
+#!/usr/bin/env bash
+
+ENIGMA_NODE_VERSION=4.4
+ENIGMA_INSTALL_DIR=$HOME/enigma-bbs
```
NuSkooler: Suggest that the entire script is wrapped in `{` and `}`. This will ensure the whole script is complete.

misc/install.sh
NuSkooler: Nitpick: ENiGMA½ for all of it's l33tness :D

NuSkooler:
I like this overall. I do think that due to the input limitation when piping into Bash, it may be worth removing the "press a key to continue" & instead, perhaps putting in a countdown (e.g. 5-10s) and tell the user to CTRL-C if they don't wan to proceed.

If that were done, the last part could simply state something like "Finished successfully" (as it does) and instruct the user to run `./oputil config --new`

Thoughts?

davestephens:
Yep, agreed on all points. At first I thought it made sense to have the pause in case people were twitchy about what the script was doing, but I actually like non-interactive installs so I've made the tweaks discussed - as well as a few more :)

Per the commit message;
- The script is surrounded in braces to ensure the whole thing is downloaded before execution
- The pause is replaced by a 10s sleep
- The script now prompts the user to generate a config if they need it
- It's now possible to override the install path, node version and git source, e.g:
````
ENIGMA_INSTALL_DIR=~/my_directory ENIGMA_NODE_VERSION=4.4.7 ENIGMA_SOURCE=https://github.com/davestephens/enigma-bbs.git ./install.sh
````
This is probably less useful for first-timers, but good for testing stuff out from your own branch, new node versions etc.

Thoughts appreciated...or merge if you're happy!

NuSkooler:
This worked well for me! I executed as such:
`curl -o- https://...long_github_raw_path/install.sh | bash`

Merging in!

PR #92, opened 2016-09-01
Some installer script tweaks

davestephens:
Spotted a mistake in the script - if ````npm install```` failed the script didn't exit accordingly. I've also added a check for a Python executable - on Ubuntu Server for the RPi it wasn't there as part of the default build.


NuSkooler:
Excellent, thank you!

PR #93, opened 2016-09-01
Small typo in username validation error message

davestephens:
Fixed a small typo that popped up if a username was already in use.


NuSkooler:
Ty!

PR #95, opened 2016-09-03
Switch to xxhash

davestephens:
xxhash works nicely on the RPi3, have switched the mci cache over to that. Only slightly different things are that xxhash wants a seed, so I used the nice example from the docs :-) The input also needs to be a buffer...

Have checked through the trace logs, and it's definitely reusing the mci cache when loading artwork....and it works on the RPi! :+1:


davestephens:
Ignore this, farmhash got fixed! See PR #96 instead.

PR #96, opened 2016-09-03
Upgrade farmhash to v1.2.1 to fix RPi3 native extension compile errors

davestephens:
Updates to farmhash to v1.2.1 which includes the RPi3 fix.


NuSkooler:
Caught me right before I brought in the PR. Awesome!!!

PR #113, opened 2017-06-01
Install note corrections / additional info

snazzware:
"--new" is just "new" currently (guessing it was --new previously), also added an FYI about needing to relog if node wasn't already installed.


NuSkooler:
@snazzware Perfect, thanks!

Issue #115, opened 2017-06-22
Exporting FTN Bundle called 00000000.thtrue instead of 00000000.th1


apamment:
For :bug: bug reports, please fill out the information below plus any additional relevant information. If this is a feature request, feel free to clear the form.

**Short problem description**
Exporting FTN Bundle called 0000000.thtrue instead of 00000000.th1
**Environment**
- [X] I am using Node.js v6.x or higher
- [X] `npm install` reports success
- Actual Node.js version (`node --version`): v6.10.2
- Operating system (`uname -a` on *nix systems): DragonFly freya.andrew.homeunix.org 4.8-RELEASE DragonFly v4.8.0-RELEASE #4: Sun Mar 26 20:55:22 EDT 2017     root@www.shiningsilence.com:/usr/obj/home/justin/release/4_8/sys/X86_64_GENERIC  x86_64
- Revision (`git rev-parse --short HEAD`): d92b8e6

**Expected behavior**
bundle should be appended with .th1 (or th0, th2, th3 etc...)

**Actual behavior**
bundle appended with thtrue

**Steps to reproduce**
Set up an Enigma BBS on thursday as a point on fsxnet then write a mail on fsxnet.

NuSkooler:
I'm very perplexed why I've never seen this. Fix looks good though.

apamment:
Yes, I don't know why it wouldn't have happened before, unless something changed in async?

PR #116, opened 2017-06-22
Fixes thtrue in #115

apamment:
Seems to work, you will probably want to check it over though, as I don't really know what I am doing ;)


NuSkooler:
You did. I need sleep. Haha!

PR #128, opened 2017-10-27
Support npm start

danger89:
Fill-in the missing pieces in the package.json file. From now on `npm start` just works like all other nodejs apps.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,9 +4,8 @@
   "description": "ENiGMA½ Bulletin Board System",
   "author": "Bryan Ashby <bryan@l33t.codes>",
   "license": "BSD-2-Clause",
-  "main": "./core/bbs",
   "scripts": {
-    "start": "node main"
+    "start": "node main.js"
   },
   "repository": {
     "type": "git",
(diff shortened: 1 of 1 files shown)


NuSkooler:
@danger89 Thanks!

Question / comment on this: Would you mind adding explicit `.js` to the module names? Also, should `main` not point to `server.js` also?

danger89:
@NuSkooler Well, no .js is not required. So also Ghost:
https://github.com/TryGhost/Ghost/blob/master/package.json#L21

And main should point to the core node module ID:
https://docs.npmjs.com/files/package.json#main

NuSkooler:
@danger89 I realize `.js` is not required; I prefer them to be explicit in the code base.

If main is not required for `npm start`, I'd say leave it out then.

danger89:
Done

NuSkooler:
Thank you very much!

danger89:
I try to setup my own bbs board from the beginning. I found enigma bbs pretty nice. Although it's not alway that easy to join with a limited telnet ansi support. For example mystic bbs can switch to a good working ascii terminal. Also I'm missing a chat feature. And an admin configure interface or something (just like mystic - cfg). Last but not least generate a default configuration file with all options. 

Keep the good job going!

NuSkooler:
Awesome! Enig is certainly more "hands on" in a lot of areas at this stage for sure. As you run into issues like described above please feel free to enter or +1 issues in the tracker. I'd like this to be usable for people besides just myself!

PR #129, opened 2017-10-27
Show it's possible to change the telnet default port

danger89:
In my case the default port (8888) was already in use. However, it was very hard to know (more guessing) the port change in possible via telnet: { port: xxxx }.


NuSkooler:
@danger89 Thank you! Documentation on this stuff is definitely severely lacking in a lot of areas :)

PR #135, opened 2017-10-29
Add some extra begginers info

danger89:
I found this by digging into the code itself, since again oputil doesn't provide anything useful yet.


danger89:
true I didn't saw this, was better to add this higher in the manual. Instead of and the end, like.. btw.. the first user becomes sysop.

danger89:
Done

PR #138, opened 2017-11-02
CombatNet support!

davestephens:
Native CombatNet support for Enig!

First proper stab (no pun intended) at any Node.js, so be nice ;-) Works nicely on my test rig, gonna get it set up on fORCE9 tomorrow...


NuSkooler:
Hell yes, this looks great! Since this brings in RLogin, maybe I'll update one of the other door modules (I forget which) that uses RLogin as well... cuz I did it manually :D

PR #140, opened 2017-11-23
Mega rejig!

davestephens:
I've rejigged a few paths of what lives in where with the idea of it being more Docker friendly.

This looks like a massive change, but it's not really. I'll run through each thing I've changed and why:

- Created `/config` and moved `menu.hjson`, `prompt.hjson` and set the default config creation/load path to the same path.
    - This moves all config-y type stuff to the same place, meaning you can mount `/config` as a volume and Enig will load these directly. (Right now the Docker image grabs the files from the volume and injects them into the right place within the Enig directory structure. 

- Moved `/mods/art` -> `/art/general`, moved `/mods/themes` -> `/art/themes`, moved actual mods within `/mods` into `/mods/system`.
    - Bit of an OCD rejig :-) Also means the "non-theme" art can be directly mounted as a Docker volume, as can themes. 

- Created `mods/user`
    - Can be mounted as a /mods volume with Docker so a user can add their own custom stuff. This has been added to getModulePaths, but will probably need a little work and testing so that menu_util.js works cleanly with it (ie bits where Config.paths.mods is referenced).


NuSkooler:
Nice work! I've only skimmed through the changes so far, but overall this sounds good. With such big changes, I'll have to do some heavy testing. 

Some additional comments:
* I like that `menu.hjson`, `config.hjson`, and `prompt.hjson` end up living in the same place. The system already has `--config` option -- I wonder if it should specify a directory vs a file now.
* I think the "system" modules could live in `/core` really. If they are truly system, they probably shouldn't be called a "mod". This is something I've needed to clean up for a long time. That would allow for simply `/mods` for the user mods.

Last two things off the top of my head for this to land in a main branch:
* Documentation. There isn't much, but adjustments will be made where any of these paths are made.
* Some kind of conversion tool (`oputil.js upgrade` type system would be slick)

Thoughts?

davestephens:
Glad you're up for the change, I think it'd make things less complicated overall. :-)

Addressing your points in order:
 - Yeah, agreed. This would also make the code that locates config and prompt/menu.hjson simpler.
 - When I created the CombatNet module (now live on force9! :)) - I didn't even consider the mods directory. Are there any specific differences other than where the files live?
 - Agreed - had intended to tackle this if you were open to the change.
 - I'd also thought about this - but wasn't sure how much it should cover. It'd be reasonably easy to add in as a bash script under misc (as it'd be something that becomes obsolete in a few months). Initial list of stuff that comes to mind:
    - move config, prompt and menu.hjson into the right place
    - move themes that aren't luciano_blocktronics
    - anything else? I'd expect that most stuff would be covered by a `git pull`...

NuSkooler:
Going to merge now - looking great.

* Thinking about `oputil.js upgrade`: I think for now, there are such limited Enig users, some notes in `UPGRADE.md` will do.
* We can get the `@systemModule` stuff cleaned up after merge

PR #141, opened 2017-12-01
Add docker option

esell:
This adds the ability to create a pretty basic container to run the bbs inside of. Currently it doesn't support SSH but could be added fairly easily. It also assumes there is an existing config file that can be used.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,17 @@
+FROM ubuntu:xenial
+
+RUN \
+  apt-get update && apt-get upgrade -q -y && \
+  apt-get install -y git curl python  && \
+  curl -sL https://deb.nodesource.com/setup_6.x | bash - && \
+  apt-get install -y nodejs node-gyp && \
+  cd /opt && \
+  git clone https://github.com/NuSkooler/enigma-bbs.git && \
+  cd enigma-bbs && npm install && \
+  apt autoremove -y && apt-get clean && \
+  mkdir /opt/enigma-conf
+
+EXPOSE 8888
+
+WORKDIR /opt/enigma-bbs
+CMD ./main.js --config /opt/enigma-conf/config.hjson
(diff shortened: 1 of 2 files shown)


NuSkooler:
I like where you're going, however I think @davestephens may have some of this in the works already. We just merged in a lot of changes to the configuration / layout to the `0.0.8-alpha` branch related to Docker. For example, `--config` now points to a *directory* instead of a `config.hjson` file.

Another thing I'd personally like to see is the Docker image based off of Alpine Linux as the images will be drastically smaller (probably around ~65 megs or so I'd guess).

NuSkooler:
Had a tad more time to think about this. My thought is now something along the lines of this:
* Previous statements about RiPuk's work and the recent merges in `0.0.8-alpha` still apply: Any Docker related stuff should go in that branch (which will require some minor tweaks to what you have)
* If multiple "flavors" of Dockers are really wanted, let's create a sub directory (e.g. `./misc/docker`) and within that sub directory, a directory for "flavor" of Dockerfile and it's documentation.

davestephens:
Hi @esell,

Check out the image pushed to docker hub - davestephens/enigma-bbs. It's still a WIP - but my plan is that I'll merge the Dockerfile and bits with it into Enigma-proper once I'm happier with it. It works beautifully now, but as @NuSkooler says, needs some love to bring the size down. Merging it into Enig will also enable automated Docker builds and tags to be tied to branches etc.

It's actually already based off of an Alpine derivative with the required Node stuff pre-baked in, the challenge with the size comes from requiring the relevant build tools to build custom mods that users *could* add. Remove the build tools then if someone were to include a mod with native extensions needing to be built then the image wouldn't start. 

As @NuSkooler says, jump into irc and let's talk! :-)

esell:
Hey @NuSkooler and @davestephens, thanks for the info on the work you've been doing, awesome to hear! I'll jump on IRC and chat with you guys for sure, feel free to close this guy out if you want ☺

NuSkooler:
@esell Awesome! Look forward to working with you on this. I don't want to discourage in any way - Docker support is going to be slick

PR #153, opened 2018-02-12
Messaging Fixes

davestephens:
Few messaging fixes:

* Fix message display format in private mail (since recent updates)
* Fix display of msginfo1 (message counters) - was a small typo in code and incorrect view codes in ANSI (updated ANSIs accordingly)
* Fix display of message area in various message lists
* OCD: Fix log warn about unused missing view in message list (probs more of these around but noticed this one when debugging)


NuSkooler:
woot!

Issue #181, opened 2018-04-24
Invalid Format client error after successful auth


coolacid:
**Short problem description**
After a successful login the following error occurs (IP scrubbed):

```
debug1: Authentication succeeded (keyboard-interactive).
Authenticated to 1.1.1.1 ([1.1.1.1]:8889).
debug1: channel 0: new [client-session]
debug3: ssh_session2_open: channel_new: 0
debug2: channel 0: send open
debug3: send packet: type 90
debug1: Entering interactive session.
debug1: pledge: network
ssh_dispatch_run_fatal: Connection to 1.1.1.1 port 8889: invalid format
```
Non-working ssh client: `OpenSSH_7.5p1 Ubuntu-10ubuntu0.1, OpenSSL 1.0.2g  1 Mar 2016`
Working ssh client: `OpenSSH_7.2p2 Ubuntu-4ubuntu2.4, OpenSSL 1.0.2g  1 Mar 2016`

**Environment**
- [X] I am using Node.js v6.x or higher
- [X] `npm install` reports success
- Actual Node.js version (`node --version`): v6.14.1
- Operating system (`uname -a` on *nix systems): Ubuntu
- Revision (`git rev-parse --short HEAD`): df01e72

**Expected behavior**
Logged in user

**Actual behavior**
SSH Client Error

**Steps to reproduce**
Unknown so far

coolacid:
Oh, kay. different user, same host works fine. I doubt this is a server side issue however, going to document it here just to be sure it's "documented"

coolacid:
Found the issue - if the ssh client intends to use compression, this error happens:

Steps to reproduce:
create a file `~/.ssh/config` with the following config:

```
Host *
    Compression yes
```

Then attempt to connect.

coolacid:
Putting a link to possible upstream bug here so that it's trackable without having to go through the PR to find the link: https://github.com/mscdex/ssh2/issues/594

NuSkooler:
@coolacid Did you ever find out any more information on this? Compression is still disabled which seems to work, but is probably not ideal.

NuSkooler:
Closing due to inactivity. Many users using SSH at this point. If you're experiencing this, feel free to re-open.

PR #182, opened 2018-04-24
Force no compression

coolacid:
It appears as tho there is a problem with compression and the upstream library. This PR forces no compression mode, making a work around for #181. 

This work around is derived from https://github.com/mscdex/ssh2/issues/594 which may be the cause for #181 as well.


NuSkooler:
Seems reasonable. I'll try this out tonight and merge it in if there are no notification drawbacks.

PR #183, opened 2018-04-28
Node.js 10 Upgrade

davestephens:
* Upgrade to Node.js 10
* Upgrade Sqlite3 package to v4 (v3 won't compile with Node.js 10)
* Fix deprecation warnings (all around implementations of Buffers)

I've had a poke around and everything seems OK, but if this can be merged to 0.0.9 I'll get it built into the Docker image and spun up on fORCE9 :+1:

### DIFF
--- a/core/archive_util.js
+++ b/core/archive_util.js
@@ -78,7 +78,7 @@ module.exports = class ArchiveUtil {
 			Object.keys(Config.fileTypes).forEach(mimeType => {
 				const fileType = Config.fileTypes[mimeType];
 				if(fileType.sig) {
-					fileType.sig 	= Buffer.alloc(fileType.sig, 'hex');
+					fileType.sig 	= Buffer.from(fileType.sig, 'hex');
 					fileType.offset	= fileType.offset || 0;
 
 					//	:TODO: this is broken: sig is NOT this long, it's sig.length long; offset needs to allow for -negative values as well
--- a/core/archive_util.js
+++ b/core/archive_util.js
@@ -82,7 +82,7 @@ module.exports = class ArchiveUtil {
 					fileType.offset	= fileType.offset || 0;
 
 					//	:TODO: this is broken: sig is NOT this long, it's sig.length long; offset needs to allow for -negative values as well
-					const sigLen =fileType.offset + fileType.sig.length;
+					const sigLen = fileType.offset + fileType.sig.length;
 					if(sigLen > this.longestSignature) {
 						this.longestSignature = sigLen;
 					}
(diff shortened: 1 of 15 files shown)


NuSkooler:
Awesome! This is something I've really wanted to look into - glad you beat me to it! First question: Are there any incompatible issues that y you're aware of with 8.x LTS? (https://github.com/nodejs/Release#release-schedule) Upgrading to the latest shiny is harder for those running ARM systems and the like it seems.

core/fnv1a.js
```diff
@@ -38,7 +38,7 @@ module.exports = class FNV1a {
 
 	digest(encoding) {
 		encoding = encoding || 'binary';
-		let buf = new Buffer(4);
+		let buf = Buffer.alloc(4);
```
NuSkooler: I should have probably been using a const here. FWIW, I think we can use `allocUnsafe()` in cases such as this where we *know* we're about to overwrite the entire chunk of memory anyway (https://nodejs.org/api/buffer.html#buffer_buffer_from_buffer_alloc_and_buffer_allocunsafe)

package.json
```diff
@@ -56,6 +56,6 @@
   },
   "devDependencies": {},
   "engines": {
-    "node": ">=6.9.2"
+    "node": ">=10"
```
NuSkooler: RE: the previous question, I'm *hoping* we can keep this 8.x LTS+... I'm just not sure yet.

davestephens:
I've pushed a few fixes - I knew I'd miss one or two allocs vs froms! 

I left the allocs as allocs rather than allocUnsafe as I figured the speed difference would be negligible, and the flip-side would likely be one of those bugs that goes unnoticed until something really weird happens at some point down the line due to memory not being cleared before being used  ;-)

RE Node.js 10 - I need to test it on my Pi3. It's the LTS-to-be (as I'm sure you're aware) - so figured it wouldn't hurt jumping the gun in setting it as a requirement - assuming it works everywhere of course!!

NuSkooler:
@davestephens Sweet, I'll try out this branch with the latest 8.x and see how it goes as well. It would be a bonus if it worked, or if it was trivial to make work with both, but if not I agree moving to 10.x is still the way to go

core/archive_util.js
```diff
@@ -78,11 +78,11 @@ module.exports = class ArchiveUtil {
 			Object.keys(Config.fileTypes).forEach(mimeType => {
 				const fileType = Config.fileTypes[mimeType];
 				if(fileType.sig) {
-					fileType.sig 	= new Buffer(fileType.sig, 'hex');
+					fileType.sig 	= Buffer.alloc(fileType.sig, 'hex');
```
NuSkooler: This one needs to use `.from()`

davestephens:
OK, I think we're there...

NuSkooler:
Wahoo!

davestephens:
Wah wah we wah!

PR #199, opened 2018-06-22
Tabs to spaces 0.0.9 alpha

NuSkooler:
ENiGMA 1/2 will now use spaces (4) instead of tabs. I personally prefer tabs, but GitHub displays them crazy, so I've caved.

Updated ESLint rule & `--fix` files. There is probably some touch up from here, but this is the bulk.


SemperFu:
https://www.youtube.com/watch?v=SsoOG6ZeyUI

NuSkooler:
I much prefer tabs, but in the end it doesn't really matter.

PR #201, opened 2018-07-16
'merge' config - merge sysop entries <30m apart

richinseattle:


core/last_callers.js
```diff
@@ -74,11 +74,25 @@ exports.getModule = class LastCallersModule extends MenuModule {
                                     return false === User.isRootUserId(parseInt(lh.log_value)); //  log_value=userId
                                 });
 
+                                var prevUserId = 0;
```
NuSkooler: Pretty much never need to use `var` anymore - use `let`.

NuSkooler:
I don't mind bringing this in, but I think it needs a little cleanup first & seems op login hiding & merging near N should be separate concepts.

core/last_callers.js
NuSkooler: Closing this since another system was implemented. If it's missing something that was here, lemme know!

PR #205, opened 2018-11-18
Add Windows pipe to Bunyan comment

SemperFu:


misc/config_template.in.hjson
```diff
@@ -66,7 +66,8 @@
         //  See https://github.com/trentm/node-bunyan#streams
         //
         //  Remember you can pipe logs through Bunyan to pretty-print:
-        //  tail -F ./logs/enigma-bbs.log | bunyan
+        //  Linux   : tail -F ./logs/enigma-bbs.log | bunyan
+        //  Windows : Get-Content .\enigma-bbs.log  | bunyan.cmd
```
NuSkooler: This should probably read "PowerShell" (which is available in Linux also)

SemperFu: Fair enough, Also added tail and -f types.

PR #228, opened 2018-12-29
Update docker.md

frozenfoxx:
Updated the command to reflect current syntax. The previously-listed version had a backslash in the container reference which caused a failure to run. The `:latest` tag is added for a reminder to the user that it is going to pull the latest image.


NuSkooler:
Thanks! I have on my To-Do to spend some time giving the Docker version some love. Any help is greatly appreciated!

PR #231, opened 2019-02-14
Update sample command for generating RSA key

frozenfoxx:
Updated the OpenSSL command to use aes128 over des3.

### DIFF
--- a/core/config.js
+++ b/core/config.js
@@ -218,7 +218,7 @@ function getDefaultConfig() {
 				//	Private key in PEM format
 				//
 				//	Generating your PK:
-				//	> openssl genrsa -aes256 -out ./config/ssh_private_key.pem 2048
+				//	> openssl genrsa -aes128 -out ./config/ssh_private_key.pem 2048
 				//
 				//	Then, set servers.ssh.privateKeyPass to the password you use above
 				//	in your config.hjson
--- a/core/config.js
+++ b/core/config.js
@@ -218,7 +218,16 @@ function getDefaultConfig() {
 				//	Private key in PEM format
 				//
 				//	Generating your PK:
-				//	> openssl genrsa -aes128 -out ./config/ssh_private_key.pem 2048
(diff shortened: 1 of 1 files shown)


NuSkooler:
Good call 👍 

Mind adding a little blurb / example of all three+ options and as to why one might choose them (or at least links to...). All technically work, some are more secure, some clients may not support all options, etc.

frozenfoxx:
Agreed and done. The comment now better describes the options available. I think we should probably update the wiki though to better explain in depth. After it's been updated we can provide a URL in the file linking to the Wiki page. It would also be good for more detailed troubleshooting info since the only reason I discovered aes256 may have problems was on startup I received an error about the password being incorrect but 3des and aes128 did not have that problem.

In any event though it's definitely a lot clearer now, tested out on my Docker container config and it immediately terminates SSH connections now without issue.

rjmooney:
FWIW, I wouldn't even make 3DES an option. It's pretty weak and has been deprecated for a number of years.

NuSkooler:
@rjmooney While using deprecated suites sucks, recall this:
https://github.com/NuSkooler/enigma-bbs/commit/25560cb47a9f7d015465414e8dfafe5aa9fc4939#diff-5324fa409c431d2a45b6e9578b78b87b

I'll do some testing with SyncTERM and NetRunner's SSH implementations -- both use cryptlib, and part of the fix was to enable 3des-cbc for SyncTERM to function at all :(

NuSkooler:
OK, did some more digging:

- `ssh2` which ENiG uses does not support aes192 or aes256 currently it seems
- Since -<cipher> is only used for the encryption of the key, not a huge deal
- OpenSSL has semi-deprecated `genrsa` in favor of the much more robust `genpkey`

I updated the SSH docs, and the example in `config.js` and [the config template](https://github.com/NuSkooler/enigma-bbs/blob/master/misc/config_template.in.hjson#L112) in both master and 0.0.10-alpha accordingly. All of them now provide the `-aes128` RSA generation form.

PR #235, opened 2019-05-30
0.0.10 alpha

NuSkooler:
Hope you don't mind, I wanna see some of this stuff. Makes it easier to see the changes anyway.


core/servers/chat/mrc_multiplexer.js
```diff
@@ -0,0 +1,269 @@
...
+                data.toString().split('\n').forEach( item => {
+                    if (item == '') return;
+
+                    // save username with socket
+                    if(item.startsWith('--DUDE-ITS--')) {
+                        connectedSockets.add(socket);
+                        socket.username = item.split('|')[1];
+                        Log.debug( { server : 'MRC', user: socket.username } , 'User connected');
+                    } else {
+                        self.receiveFromClient(socket.username, item);
```
NuSkooler: This is the only ref to `self` above. Should be able to just use `this` by using arrow function above vs `function(socket)`

core/servers/chat/mrc_multiplexer.js
```diff
@@ -0,0 +1,269 @@
...
+            // if not a heartbeat, and we have clients then we need to send something to them
+            this.sendToClient(message);
+        }
+    }
+
+    /**
+     * Takes an MRC message and parses it into something usable
+     */
+    parseMessage(line) {
+        const msg = line.split('~');
```
NuSkooler: FWIW you can do unpacking stuff like:
```
const [from_user, from_site, from_room, ... ] = line.split(~);
// ...
return { from_user, from_site, from_room, ... }; // also short syntax
```

core/servers/chat/mrc_multiplexer.js
```diff
@@ -0,0 +1,269 @@
...
+
+/**
+ * SLugifies the BBS name for use as an MRC "site name"
+ */
+function slugify(text) {
+    return text.toString()
+        .replace(/\s+/g, '_')           // Replace spaces with _
+        .replace(/[^\w\-]+/g, '')       // Remove all non-word chars
+        .replace(/\-\-+/g, '_')         // Replace multiple - with single -
+        .replace(/^-+/, '')             // Trim - from start of text
```
NuSkooler: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/trim

core/servers/chat/mrc_multiplexer.js
```diff
@@ -0,0 +1,269 @@
...
+        this.server = net.createServer( function(socket) {
+            socket.setEncoding('ascii');
+
+            socket.on('data', data => {
+                // split on \n to deal with getting messages in batches
+                data.toString().split('\n').forEach( item => {
+                    if (item == '') return;
+
+                    // save username with socket
+                    if(item.startsWith('--DUDE-ITS--')) {
```
NuSkooler: Oh man, that protocol! haha

core/config.js
```diff
@@ -447,6 +449,23 @@ function getDefaultConfig() {
             }
         },
 
+        chatServers : {
+            mrc: {
+                enabled             : true,
+                serverHostname      : 'mrc.bottomlessabyss.net',
+                serverPort          : 5000,
+                multiplexerPort     : 5000,
+                bbsInfo : {
```
NuSkooler: You can get +op name from `StatLog.getSystemStat(SysProps.SysOpUsername)` or `StatLog.getSystemStat(SysProps.SysOpRealName)`

I wonder if some of those properties should be part of the general configuration along with `boardName` et. al.

core/servers/chat/mrc_multiplexer.js
```diff
@@ -0,0 +1,325 @@
...
+
+//  deps
+const net                   = require('net');
+const _                     = require('lodash');
+const os                    = require('os');
+
+
+// MRC
+const protocolVersion   = '1.2.9';
+const lineDelimiter     = new RegExp('\r\n|\r|\n');
```
NuSkooler: `string_util.js` has `splitTextAtTerms` you could use

core/servers/chat/mrc_multiplexer.js
```diff
@@ -0,0 +1,325 @@
...
+        this.mrcConnectOpts = {
+            host         : this.config.chatServers.mrc.serverHostname || 'mrc.bottomlessabyss.net',
+            port         : this.config.chatServers.mrc.serverPort || 5000,
+            retryDelay   : this.config.chatServers.mrc.retryDelay || 10000
+        };
+    }
+
+    _connectionHandler() {
+        const config = Config();
+        const boardName  = config.general.prettyBoardName || config.general.boardName;
```
NuSkooler: What's a pretty board name? (just curious)

core/servers/chat/mrc_multiplexer.js
```diff
@@ -0,0 +1,325 @@
...
+};
+
+const connectedSockets = new Set();
+
+exports.getModule = class MrcModule extends ServerModule {
+    constructor() {
+        super();
+
+        this.log            = Log.child( { server : 'MRC' } );
+        this.config         = Config();
```
NuSkooler: I _think_ this is only used locally, so probably just want `const config = Config()`

core/servers/chat/mrc_multiplexer.js
```diff
@@ -0,0 +1,325 @@
...
+            host         : this.config.chatServers.mrc.serverHostname || 'mrc.bottomlessabyss.net',
+            port         : this.config.chatServers.mrc.serverPort || 5000,
+            retryDelay   : this.config.chatServers.mrc.retryDelay || 10000
+        };
+    }
+
+    _connectionHandler() {
+        const config = Config();
+        const boardName  = config.general.prettyBoardName || config.general.boardName;
+        const enigmaVersion = 'ENiGMA-BBS_' + require('../../../package.json').version;
```
NuSkooler: Can their servers handle `ENiGMA½`? It fits within CP437.

core/servers/chat/mrc_multiplexer.js
```diff
@@ -0,0 +1,325 @@
...
+    connectToMrc() {
+        const self = this;
+
+        // create connection to MRC server
+        this.mrcClient = net.createConnection(this.mrcConnectOpts, self._connectionHandler.bind(self));
+
+        this.mrcClient.requestedDisconnect = false;
+
+        // do things when we get data from MRC central
+        var buffer = new Buffer.from('');
```
NuSkooler: `let buffer = ...`

core/servers/chat/mrc_multiplexer.js
```diff
@@ -0,0 +1,325 @@
...
+
+        function handleData(chunk) {
+            if (typeof (chunk) === 'string') {
```
NuSkooler: ENiG requires & uses lodash all of the the place, so you can e.g. 
```
const _ = require('lodash');
// ...
if(_.isString(chunk)) { ... }
```

core/servers/chat/mrc_multiplexer.js
```diff
@@ -0,0 +1,325 @@
...
+
+            if (lines.pop()) {
+                // if buffer is not ended with \r\n, there's more chunks.
+                return;
+            } else {
+                // else, initialize the buffer.
+                buffer = new Buffer.from('');
+            }
+
+            lines.forEach(function iterator(line) {
```
NuSkooler: General comment: Prefer arrow functions
```
lines.forEach( line => { ... } )
```

core/servers/chat/mrc_multiplexer.js
davestephens: MRC needs two "formats" of your board name - one slugified which it uses as a "site name" in messages, and the "pretty" one, which can have pipe codes in and is shown when a user runs the `/BBSES` command. I didn't want to add it into `config.js` as it'd break pretty much everyone that didn't have it, but I thought a nice extra option to add for people that want to take advantage of it. What are your thoughts?

core/servers/chat/mrc_multiplexer.js
davestephens: Good spot...

core/servers/chat/mrc_multiplexer.js
davestephens: oops!

core/servers/chat/mrc_multiplexer.js
davestephens: :+1:

core/servers/chat/mrc_multiplexer.js
davestephens: oh, much nicer!

core/servers/chat/mrc_multiplexer.js
davestephens: I'm missing how `trim` is useful here?

core/servers/chat/mrc_multiplexer.js
davestephens: I can change it to `--BRO-ITS--`? :-)

core/config.js
davestephens: I think the `bbsInfo` stuff should be, yeah. I'll move it there and I'll change it if you hate it :)

core/servers/chat/mrc_multiplexer.js
NuSkooler: Nope sorry, I needed to look at the regex closer - you're removing '-', not spaces.

core/servers/chat/mrc_multiplexer.js
NuSkooler: Makes sense. I think you can add it to the config.hjson template + `config new` can populate it with the same value as `boardName` with a comment stating it can allow pipe color codes. It could be exposed and used elsewhere later (MCI code perhaps)

core/servers/chat/mrc_multiplexer.js
```diff
@@ -235,20 +238,15 @@ exports.getModule = class MrcModule extends ServerModule {
...
-            from_site: msg[1],
-            from_room: msg[2],
-            to_user: msg[3],
-            to_site: msg[4],
-            to_room: msg[5],
-            body: msg[6]
-        };
+        const [from_user, from_site, from_room, to_user, to_site, to_room, body ] = line.split('~');
+
+        // const msg = line.split('~');
```
NuSkooler: If you want your sanity check still, `body` (and maybe others) would simply be undefined if not enough elements were produced from the split.

davestephens: I think it should be ok now - it was there because I didn't have the tcp buffering/chunking in place

davestephens:
Right, so I think this is pretty much there now...!

NuSkooler:
Sweet, I'll see about getting this merged in the next day or so!

PR #236, opened 2019-06-22
MRC Fixes

davestephens:
Fixes:
- Broken `/INFO` and lack of use of `prettyBoardName`
- Reduces GUI by one line to prevent display issues


NuSkooler:
<3

PR #273, opened 2020-04-26
Update testing.md

squintgit:
Adding instructions for shutting down server, which may not be immediately clear to new sysops


NuSkooler:
ty!

PR #274, opened 2020-05-03
Qwk support

NuSkooler:
Initial QWK support exposed only with `oputil.js` until some additional testing occurs.


core/qwk_mail_packet.js
```diff
@@ -279,13 +313,54 @@ class QWKPacketReader extends EventEmitter {
...
+
+            const FTNKludgeMapping = {
+                'X-FTN-PATH'    : 'PATH',
+                'X-FTN-MSGID'   : 'MSGID',
+                'X-FTN-REPLY'   : 'REPLY',
+                'X-FTN-PID'     : 'PID',
+                'X-FTN-FLAGS'   : 'FLAGS',
+                'X-FTN-TID'     : 'TID',
+                'X-FTN-CHRS'    : 'CHRS',
+                //  :TODO: X-FTN-KLUDGE - not sure what this is?
```
rswindell, line 519: X-FTN-KLUDGE is for otherwise-unrecognized/unsupported FTN control lines.

NuSkooler: @rswindell Thanks! What is the format?

rswindell: The contents (value) of the header field is everything found between (and not including) the Ctrl-A character and the carriage-return. e.g. X-FTN-KLUDGE: SOMEKLUDGE
Often the kludge will have a colon-terminated field name, not always.

NuSkooler: @rswindell sorry for all the questions - I guess multiple `X-FTN-KLUDGE` entries are allowed?

rswindell: Yes, multiple would be allowed. It's my way of retaining all the metadata of the original message and being able to reproduce it, even if I don't recognize it.

Issue #280, opened 2020-05-17
Configuration File Revamp


NuSkooler:
Configuration files (`config.hjson`, `menu.hjson`/`prompt.hjson` and friends, `theme.hjson`, etc.) could use a revamp.

Creating this ticket to replace some existing so we have a little better organization:
* Configuration / Menu Include System #220
* Suggestion - Break up Menu file #158

Updates on current design thoughts:
## General Goals
* Remain HJSON driven
* Configurations to remain hot-reload
* Self references (reference 'a.b', from 'a.d.e', etc.)
* Templates (use self ref as a template)
* Environment variable support
* Includes
* Good reporting of syntax errors

### Includes
A the very least, a top level include:
```
// my-bbs.hjson
{
  includes: [
    // default menus
    enigma-bbs-default-menus.hjson
    // others
    my-menus.hjson
  ]
}
```

## Self References
```
{
  foo: {
    bar: {
      baz: foobar!
    }
  }
  
  fubar: {
    // resolves to { baz: foobar! } above
    bar: @reference:foo.bar
  }
}
```

## Environment Variables
```
{
  foo: {
    // SOME_VAR in current process environment
    // if no type is supplied, the value is a string
    bar: @environment:SOME_VAR
    
    // see notes on types below
    barBoolean: @environment:BOOL_VAR:boolean

    // arrays
    barNumberArray: @environment:NUM_VAR:number:array
  }
}
```
As shown above, environment variables default to strings. An optional type suffix can be supplied to coerce to a particular type (if the conversion is not possible, the full spec will remain). An additional `array` suffix can be supplied to convert to an array (think `export FOO="1,2,3"` -> `[ 1, 2, 3 ]`).

## Notes
* Main file (that is, where `includes` reside) override any configuration from included files
* For simplicity, `includes` will be main/top-level only (no nesting)

Feedback is of course very welcome! If you're like to play with a toy implementation that illustrates some of the above follow this link: https://repl.it/@NuSkooler/ConfigRevampTests

planeguy:
I looked around a bit, and there's a small community using this syntax for includes:

```json
{
    "property": { "...":"file-to-include.json"}
}
```

a thought about this is if we merge at load time (it looks like we reload the config with file watch?), then we could do double-duty on includes and self-references eg:

```
{
   a: {...: thing.hjson}
   b:{...:thing.hjson}
   c: {
     art: CONNECT
     ...:thing.hjson
   }
}
```

[...]

PR #292, opened 2020-07-07
#280 Configuration Revamp

NuSkooler:
This implements the features discussed in #280. Additionally, menu files are broken into parts when creating a default configuration with `./oputil.js config new`.

This will be beta tested on Xibalba and I welcome feedback before it's ready to merge.


core/config_loader.js
```diff
@@ -0,0 +1,240 @@
+//  deps
+const paths = require('path');
+const async = require('async');
+const moment = require('moment');
+
+const _ = require('lodash');
+const mapValuesDeep = require('deepdash/getMapValuesDeep')(_);
+
+module.exports = class ConfigLoader {
```
NuSkooler, line 9: `ConfigLoader` is a generic loader for HJSON configuration files that use includes, @references, etc.

PR #293, opened 2020-07-07
Update msg_list.js

planeguy:
index 0 is falsey so it ends up falling back to the deprecated arg name


NuSkooler:
Oops sorry I meant to pluck the fix I have in the config revamp branch. Please see there or I'll get a new update in.

PR #317, opened 2020-10-01
Minor Documentation Updates

blackmagic-consult:
Added some things that were missing in the documentation that hindered me when I was setting up my own install.

These changes are:

1. Filling out the Networking page to include information on port forwarding and DynDNS. It's sparse, but a good foot hold for people unfamiliar with the topics.
2. Mentioning that files served by the web server are available by default from the "/static/" route
3. Mentioning that CertBot requires a privilege adjustment to work with a instance running as non-route.

Thanks for the great tool, NuSkooler!

### DIFF
--- a/docs/installation/network.md
+++ b/docs/installation/network.md
@@ -2,5 +2,12 @@
 layout: page
 title: Network Setup
 ---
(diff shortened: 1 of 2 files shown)


NuSkooler:
Thank you for these updates!

Issue #259, opened 2020-03-01
Change Spacing on Acheivements


flewid:
The achievments list isn't spaced properly, after the "8:12pm" or other entries that don't have two digits, a space should be entered 'automatically' so the spacing remains the same.

see screenshot:

[link](https://www.dropbox.com/s/h007bqitq7invo1/Screenshot%202020-02-29%2023.50.04.png?dl=0)

NuSkooler:
I'll look at this when I get a chance, but basically just adjust the format code to adjust e.g. the date column min width in the `theme.hjson`. You can see other examples there. Accepting PRs!

flewid:
Great thanks. Fixed it.

dateTimeFormat: MMM Do h:mma
to
dateTimeFormat: MMM Do hh:mma

Thanks!

cognitivegears:
@NuSkooler Sounds like this was fixed quite a while ago, can this be closed?

PR #318, opened 2020-10-01
Achievement spacing

blackmagic-consult:
Edited the default theme.hjson so wherever "h:mm a" occured, it was replaced with "hh:mm a", which corrects all spacing issues in the Mystery Skull theme.

Referring to issue #259

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,6 @@
 
 ENiGMA½ is a modern BBS software with a nostalgic flair!
 
-
 ## Features Available Now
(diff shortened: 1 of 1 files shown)


blackmagic-consult:
Dammit ,I was working in a new branch - why did my documentation edits carry over >:(

NuSkooler:
@blackmagic-consult Do you want to separate the doc updates? I'm not super picky PR-wise. They look great!

blackmagic-consult:
Should be clear now! Thanks NuSkooler.

NuSkooler:
Are you sure this actually fixes the issue **in all areas**? It's setting the date/time to a fixed width, but that *available* width/alignment varies per view -- generally controlled via `itemFormat` and the like.

PR #319, opened 2020-10-01
Added an additional digit to the hour slot, fixing minor spacing issue

blackmagic-consult:
Hi, sorry, been a long time since I've had to merge on any project that wasnt on BitBucket.

Yes, I checked this against every window that renders the time in the default theme, and it adjusts the wonky spacing that exists before without stepping outside the "bounding box" for lack of a better term.

Thanks NuSkooler.


NuSkooler:
Thanks @blackmagic-consult !

Issue #333, opened 2020-11-29
Removed files not fully deleted from sqlite DB


deepthaw:
Removing files from system using ./oputil.js does not appear to fully remove all the data in the relevant tables. This results in files that are added later having potentially incorrect data due to linking to the previous files.

Quickest way to recreate: Import some files with given tags. Remove the files from the system, then import them with different tags. The old tags will still be on the freshly imported files.

deepthaw:
Can verify that at least the file_hash_tag table is not getting updated when files are removed. Ran "sqlite3 file.sqlite3" and SELECT * FROM file_hash_tag has multiple entries pointing towards files that no longer exist.

NuSkooler:
Fix + notes on manual updates merged in

PR #335, opened 2020-11-29
add cascading delete to multiple keys within file db

deepthaw:
Added cascading delete to file_id fk in multiple places. should resolve bugs related to #333. existing users will need a way to update their existing db if they don't want to delete it to have it recreated.


NuSkooler:
Thank you for this! Snagged these changes + a few more related & upgrade path to new PR which is now merged.

PR #338, opened 2020-12-06
Update abracadabra.js

deepthaw:
fix for next parameter in abracadabra. allows usage of doors in login/etc. without entering infinite loop.


NuSkooler:
Nice!

Issue #337, opened 2020-11-30
Outgoing Rlogin Gateway


deepthaw:
I'd like to see a way to create outgoing rlogin connections, including authentication strings. This would allow users to be passed transparently to another bbs, possibly functioning as a gameserver.

How Synchronet handles incoming rlogin attempts is documented at: http://wiki.synchro.net/server:terminal

This would differ from telnet in that (to my knowledge) rlogin allows more data to passed along, which can be used for authorization purposes.

deepthaw:
The existing CombatNet module is almost exactly already in a state where it can be used for exactly what I wanted. It took only minor modifications to turn it into a generic rlogin module - would it be worth pulling these back to the main branch?

NuSkooler:
I think a number of people would fine this useful, so yeah a PR would be great!

deepthaw:
Question: Should the combatnet module remain, or should it be replaced by a generic rlogin module? The necessary changes are minimal, but if a bug should crop up in one, it'll need to be fixed in the other.

NuSkooler:
@deepthaw If it can be made to work generically I think that's better. Would have to come with some document updates, but probably not thing major -- WHATSNEW/UPGRADE/etc.

NuSkooler:
Closing due to inactivity. The PR has been closed for a while as well due to the same thing.

PR #339, opened 2020-12-06
Create rlogin.js

deepthaw:


deepthaw:
This addresses #337

NuSkooler:
A couple things - 
1. This seems like it did end up being almost a copy of the CombatNet module - can that one be deprecated and this be made to support both? This would require a couple updates in the docs.
2. I want to encourage all new code to use arrow functions as much as possible, e.g.:
```js
const foo = () => { }; // arrow
// vs
const foo = function() {
};

// or callbacks:
(arg1, arg2) => {
}

// vs 
function callback(arg1, arg2) {
}
```

I'm actually plan on making `async` and `await` the real way to go. I wish I had that years go when starting this stuff.

NuSkooler:
Closing due to inactivity. Feel free to re-open with an update.

PR #379, opened 2022-01-23
New view type - Full Menu View

cognitivegears:
This started because I really liked the way the login menu was setup. I found myself wishing that I could use menus like that throughout the app, including places that had multiple columns (for example, the Main Menu). So I developed a new type of view component to handle it, the Full Menu (FM) view. This view works similarly to VM and HM (and in fact, it can replace those in most circumstances.) In addition to showing a single column, however, this also allows for the display of more than one column in a single view.

To extend it even further, I added on multi-page support. Similar to VM, it can display as list of items larger than the height/width given to it, by splitting the items into multiple "pages". This allows scrolling with up/down, as well with PageUp/PageDown, Home and End.

I also added on support for a few additional items that were not completely implemented in the others. For example, `justify` works to set the justification, and `fillChar` can be used to fill the empty spaces with a character other than a space. In addition to `itemSpacing` I added on an `itemHorizSpacing` to be able to set spacing between columns as well. Finally, I implemented `textOverflow` from other views, in the case where only a single column is displayed to control what happens to text that goes beyond the width of the view.

I tried to keep changes outside the view to a minimum, but there were a few that were needed. In `mci_view_factory.js` I added the `FM` MCI type for the view. In `menu_view.js` I added support for some of the properties described above. In `view.js` I fixed a likely issue that caused setFocus to be called twice if focused is defined in the property, the first time before all of the other properties had necessarily been set. This is mostly a performance fix to avoid having to render twice in views in that case.

Since this is a new type of view, I also added documentation for it, both so you can see more about how it works as well as for end users. I thought it might be useful in the future to document other views in the same way. Also there are a few screenshots to give an idea of the basic operation.

Please let me know if there is any additional information you need, or if there are any changes / enhancements you would like to see. Hope you like it!


NuSkooler:
@cognitivegears Thanks for this, it looks awesome! I'll read through the code in just a bit, but just skimming it looks great. I appreciate that you included the docs and examples (they're better than mine!).

cognitivegears:
Wow thanks I appreciate it. I have already learned a lot from your code, Enigma 1/2 is an impressive project! Please let me know if you have any questions or see any issues.

core/full_menu_view.js
```diff
@@ -0,0 +1,522 @@
...
+
+  // Initialize paging
+  this.pages = [];
+  this.currentPage = 0;
+
+  this.initDefaultWidth();
+
+
+
+  const self = this;
```
NuSkooler, line 33: For newer Enig code you can use ES6. This means you can use arrow functions (`=>`) that respect what you're used to for `this`... so no need to capture a `self`, either.

core/full_menu_view.js
```diff
@@ -0,0 +1,522 @@
...
+
+
+
+  this.getSpacer = function() {
+    return new Array(self.itemHorizSpacing + 1).join(this.fillChar);
+  }
+
+
+  this.clearPage = function() {
+    var width = this.dimens.width;
```
NuSkooler, line 62: Please try to use `let` and `const`, no `var` -- it's very rarely needed.

docs/art/views/full_menu_view.md
```diff
@@ -0,0 +1,231 @@
+---
```
NuSkooler, line 1: These docs are great!

cognitivegears: Thanks, once this PR is done I was thinking I'd make similar docs for the other views so it will be consistent.

core/full_menu_view.js
cognitivegears: I believe resolved. I didn't go all the way to convert to an ES6 class (might be cool to think about in the future, but would impact the full object hierarchy) but let me know if what I changed is good for now.

core/full_menu_view.js
cognitivegears: Updated. How embarrassing :) It's been years since I've done a lot of serious JS work and old habits die hard. lol

docs/art/views/full_menu_view.md
NuSkooler: Oh man that would be great. I've put a lot of work into the docs, but as you can see they still need a lot more love!

core/full_menu_view.js
NuSkooler: Looks great! Some of it is a bit tricky due to the code base. When I started, ES6 wasn't quite ready, so you can see this evolution in the code. You'll see this around arrow functions as well. Async/await is not available also and I'd like to start producing more APIs to support it to make the code cleaner, but that's how laundry lists just keep getting longer :D

core/full_menu_view.js
NuSkooler: Totally get it. It's hard when you bounce around in different runtimes too

NuSkooler:
@cognitivegears I'm happy to merge this in when you give me the go ahead -- I don't want to merge it in if you're still wanting to make any minor changes.

Issue #382, opened 2022-01-24
docker implementation fails when /enigma-bbs/art is mounted to an external folder


abutbul:
**Describe the Bug**
this bug happens to me while configuring and running enigma for the first time. while using docker set-up instructions as documented in this project and with the image directly from dockerhub. combining with my own changes (partly in order to fix the documentation which currently lacks db persistency)

**To Reproduce**
steps 
```
docker run --rm -it \
-v "{HOME}/enigma-bbs/config:/enigma-bbs/config" \
-v "{HOME}/enigma-bbs/db:/enigma-bbs/db" \
-v "{HOME}/enigma-bbs/logs:/enigma-bbs/logs" \
-v "{HOME}/enigma-bbs/filebase:/enigma-bbs/filebase" \
-v "{HOME}/enigma-bbs/art:/enigma-bbs/art" \
enigmabbs/enigma-bbs:latest \
oputil.js config new
```
configure your board and now start it (due to --rm - was used to make sure no residue is left - along with prune before every test)
```
docker run --rm -d -p 8888:8888 \
-v "/share/Container/enigma-bbs/config:/enigma-bbs/config" \
-v "/share/Container/enigma-bbs/db:/enigma-bbs/db" \
-v "/share/Container/enigma-bbs/logs:/enigma-bbs/logs" \
-v "/share/Container/enigma-bbs/filebase:/enigma-bbs/filebase" \
-v "/share/Container/enigma-bbs/art:/enigma-bbs/art" \
enigmabbs/enigma-bbs:latest
```

**Expected Behavior**
bbs should accept connections in 8080

**Actual Behavior**

bbs is unable to serve connections due to missing files in art folder (see environment section)

**Screenshots**
error message
```
{"name":"ENiGMA½ BBS","hostname":"23814a6ec4dd","pid":19,"nodeId":1,"sessionId":"vnUpYvxbzbn","level":30,"remoteAddress":"::ffff:10.100.102.58","serverName":"Telnet","isSecure":false,"msg":"Client connected","time":"2022-01-24T22:00:06.500Z","v":0}
[...]

PR #386, opened 2022-01-25
fix docker implementation

abutbul:
fixes https://github.com/NuSkooler/enigma-bbs/issues/382
added entrypoint to make sure everything works
some file redundancy in place - minimal and needed for portability


NuSkooler:
@davestephens

docker/Dockerfile
```diff
@@ -1,32 +1,43 @@
...
-RUN apt-get update && apt-get install -y \
+# Do some installing! (and alot of cleaning up) keeping it in one step for less docker layers
+# - if you need to debug i recommend to break the steps with individual RUNs)
+RUN apt-get update \
+    && apt-get install -y \
     git \
     curl \
     build-essential \
     python \
+    python 3 \
```
davestephens, line 17: This looks suspect?

abutbul: pm ERR! code 1                                                                                                                                                                                      npm ERR! path /enigma-bbs/node_modules/xxhash                                                                                                                                                        npm ERR! command failed                                                                                                                                                                              npm ERR! command sh -c node-gyp rebuild                                                                                                                                                              npm ERR! gyp info it worked if it ends with ok
npm ERR! gyp info using node-gyp@8.3.0                                                                                                                                                               npm ERR! gyp info using node@16.13.2 | linux | x64                                                                                                                                                   npm ERR! gyp ERR! find Python                                                                                                                                                                        npm ERR! gyp ERR! find Python Python is not set from command line or npm configuration
npm ERR! gyp ERR! find Python Python is not set from environment variable PYTHON                                                                                                                     npm ERR! gyp ERR! find Python checking if "python3" can be used                                                                                                                                      npm ERR! gyp ERR! find Python - "python3" is not in PATH or produced an error                                                                                                                        npm ERR! gyp ERR! find Python checking if "python" can be used                                                                                                                                       npm ERR! gyp ERR! find Python - executable path is "/usr/bin/python"                                                                                                                                 npm ERR! gyp ERR! find Python - version is "2.7.16"
npm ERR! gyp ERR! find Python - version is 2.7.16 - should be >=3.6.0                                                                                                                                npm ERR! gyp ERR! find Python - THIS VERSION OF PYTHON IS NOT SUPPORTED                                                                                                                              npm ERR! gyp ERR! find Python                                                                                                                                                                        npm ERR! gyp ERR! find Python **********************************************************                                                                                                             npm ERR! gyp ERR! find Python You need to install the latest version of Python.                                                                                                                      npm ERR! gyp ERR! find Python Node-gyp should be able to find and use Python. If not,                                                                                                                npm ERR! gyp ERR! find Python you can try one of the following options:                                                                                                                              npm ERR! gyp ERR! find Python - Use the switch --python="/path/to/pythonexecutable"                                                                                                                  npm ERR! gyp ERR! find Python   (accepted by both node-gyp and npm)                                                                                                                                  npm ERR! gyp ERR! find Python - Set the environment variable PYTHON                                                                                                                                  npm ERR! gyp ERR! find Python - Set the npm configuration variable python:                                                                                                                           npm ERR! gyp ERR! find Python   npm config set python "/path/to/pythonexecutable"
npm ERR! gyp ERR! find Python For more information consult the documentation at:
npm ERR! gyp ERR! find Python https://github.com/nodejs/node-gyp#installation
npm ERR! gyp ERR! find Python **********************************************************
npm ERR! gyp ERR! find Python
npm ERR! gyp ERR! configure error
npm ERR! gyp ERR! stack Error: Could not find any Python installation to use
npm ERR! gyp ERR! stack     at PythonFinder.fail (/usr/local/lib/node_modules/npm/node_modules/node-gyp/lib/find-python.js:330:47)
npm ERR! gyp ERR! stack     at PythonFinder.runChecks (/usr/local/lib/node_modules/npm/node_modules/node-gyp/lib/find-python.js:159:21)
npm ERR! gyp ERR! stack     at PythonFinder.<anonymous> (/usr/local/lib/node_modules/npm/node_modules/node-gyp/lib/find-python.js:266:16)
npm ERR! gyp ERR! stack     at PythonFinder.execFileCallback (/usr/local/lib/node_modules/npm/node_modules/node-gyp/lib/find-python.js:297:7)
npm ERR! gyp ERR! stack     at ChildProcess.exithandler (node:child_process:388:7)
npm ERR! gyp ERR! stack     at ChildProcess.emit (node:events:390:28)
npm ERR! gyp ERR! stack     at maybeClose (node:internal/child_process:1064:16)
npm ERR! gyp ERR! stack     at Process.ChildProcess._handle.onexit (node:internal/child_process:301:5)
npm ERR! gyp ERR! System Linux 5.10.60-qnap
npm ERR! gyp ERR! command "/usr/local/bin/node" "/usr/local/lib/node_modules/npm/node_modules/node-gyp/bin/node-gyp.js" "rebuild"
npm ERR! gyp ERR! cwd /enigma-bbs/node_modules/xxhash
npm ERR! gyp ERR! node -v v16.13.2
npm ERR! gyp ERR! node-gyp -v v8.3.0
npm ERR! gyp ERR! not ok

abutbul: error cleared after adding py3.

abutbul: that is a great catch thank you - this is what i get for making ad-hoc changes

docker/Dockerfile
```diff
@@ -42,4 +53,4 @@ EXPOSE 8888
 
 WORKDIR /enigma-bbs
 
-CMD ["pm2-runtime", "main.js"]
+ENTRYPOINT ["/enigma-bbs/docker/bin/docker-entrypoint.sh"]
```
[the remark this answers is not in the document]
abutbul: well pm2-runtime is looping when the application crashes, starting the application with no configuration causes infinite loop which causes great pain for docker users. the entry point fixes it, and inits the volumes where needed.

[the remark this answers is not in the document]
abutbul: also - give it a spin see how it behaves
i tested several scenarios
including no volumes, some volumes, partial configuration
all should yield with proper error or spin up a working instance.

NuSkooler:
> This is a cool piece of functionality - I do wonder however if it's better handled within Enigma itself rather than using bash scripts. Ie, docker build could create the staging area idea you came up with, but then Enigma is the bit that copies stuff around if it detects files missing that it needs. It also means that if the base themes are changed (which happens occasionally) that too could be handled by Enigma, whereas with this approach it would be a manual copy to get the missing files in the right places...

I have to be honest, I'd rather ENiGMA 1/2 itself not have to try to cater to running in a container or not -- this should be transparent.

My understanding of what needs to be addressed here. Correct me if I'm wrong -- I don't have ton of time to go through this stuff right now:
* At first run, configuration files are not present in volume location
* At first run, art files may not be present in volumes

So the idea here is to config & copy files to their volumes, yes? Can we make this an **explicit** run option to the container _and_ prompt to be double-sure this is what users want to do?

For updates, perhaps an explicit update option would only copy over files that do not yet exist. Anything beyond that requires users to diff/merge/add but as menu files are broken up much more in recent versions this should be less impactful.

abutbul:
> > This is a cool piece of functionality - I do wonder however if it's better handled within Enigma itself rather than using bash scripts. Ie, docker build could create the staging area idea you came up with, but then Enigma is the bit that copies stuff around if it detects files missing that it needs. It also means that if the base themes are changed (which happens occasionally) that too could be handled by Enigma, whereas with this approach it would be a manual copy to get the missing files in the right places...
> 
> I have to be honest, I'd rather ENiGMA 1/2 itself not have to try to cater to running in a container or not -- this should be transparent.
> 
> My understanding of what needs to be addressed here. Correct me if I'm wrong -- I don't have ton of time to go through this stuff right now:
> 
> * At first run, configuration files are not present in volume location
> * At first run, art files may not be present in volumes

DA> i have expanded on all scenarios i could think of down below
> 
> So the idea here is to config & copy files to their volumes, yes? Can we make this an **explicit** run option to the container _and_ prompt to be double-sure this is what users want to do?
> 
DA> script will never copy over existing files currently no explicit way to tell it to do so, and i believe it should stay so and be left for more advanced docker users if they chose to stick with this solution. 

> For updates, perhaps an explicit update option would only copy over files that do not yet exist. Anything beyond that requires users to diff/merge/add but as menu files are broken up much more in recent versions this should be less impactful.

DA> this is my first version with enigma, I will look out for the upgrade and see if I can get list of relevant changes and pre-populate them in some migration script to be ran afterwards (we are keeping the assets and we can chose what to bring later on when starting the new version. script will also need a .version file somewhere inside the docker (or a place where its available) so it can check if migration is even needed. - i think it's out of scope for the current change.

in detail:

I just want to make one thing clear, the entry-point script will not overwrite any volume under any circumstance. 
scenarios and examples:

**you've started the image for the first time - with no volumes** - you will be skipped to configuration since files are in place and no default config file found

**you've configured it once (all still in temporary docker volumes) and now trying to run it without external partitions** - your configuration is saved - untouched

**you've configured it once (all still in temporary docker volumes) and now trying to run it with external volumes** - your original volumes will be untouched, your provided volumes in case they are empty will be populated with needed files and config will run. - your BBS data is still retrievable from docker volumes 
at anytime you can copy these volumes to folders and migrate to persistent config with providing the new folders.


**if you've provided an empty mount - file copy will happen - nothing exists so nothing overwrites** - worst case scenario you get a new BBS and have to check your starting folder or volume assignment

**if you've provided existing mounts (and they have files in them, any files, regardless of who put them there)** no copy will happen and configuration file logic will reoccur to see if BBS should start.


again i will stress - **entrypoint script should never will overwrite any data anywhere**. but people need to understand that without providing a volume their deployment is not persistent (unless these volumes are NEVER pruned while container is not attached - and we cant force this so we just do our best to not interrupt)

imo other than upgrade, next step is to wrap it all with a docker-compose file with possibly an .env file to make all of the volume mounts easier to read and set for docker-compose users.

NuSkooler:
Thanks for the rundown, I like it -- 

Have you looked at the config's environment variable support? https://github.com/NuSkooler/enigma-bbs/blob/master/docs/configuration/config-files.md#environment-variables

> again i will stress - entrypoint script should never will overwrite any data anywhere.

I'm still tossing around explicit vs implicit init mode... Definitely do not want to goof and wipe out someones setup.

>  but people need to understand that without providing a volume their deployment is not persistent (unless these volumes are NEVER pruned while container is not attached - and we cant force this so we just do our best to not interrupt)

We could spit out a warning to `stdout` about this at startup to make it clear

abutbul:
there are current warnings for existing files, we can make them fatal to stop the process. I didn't use any logging facility just plain text with echo.

i totally agree we can not guarantee persistency if no volumes are mounted. 

this may be something obvious to me as it is to other experienced docker users, but may not be the case for newbies. it should be stressed - perhaps we can even place some .ans welcome screen explicitly reporting that the state of deployment is on docker mounts rather than volumes. 

if some one past that welcome screen and said to himself, let's get hackin on this board, he may experience many other issues along the way apart for volume prune killing his files. 



we can wrap around the docker run command by some questionnaire to make sure the param are ok - but i haven't encountered many a project that went down this road.

abutbul:
> Thanks for the rundown, I like it --
> 
> Have you looked at the config's environment variable support? https://github.com/NuSkooler/enigma-bbs/blob/master/docs/configuration/config-files.md#environment-variables
> 
missed answering that one, it's pretty neat. how do you see it being used in this flow?

docker/bin/docker-entrypoint.sh
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env bash
+
+# Set some vars
+prepopvols=("config" "mods" "art") # these are folders which contain runtime needed files, and need to be represented in the host
```
NuSkooler, line 4: Perhaps `set -e` ?

docker/bin/docker-entrypoint.sh
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env bash
+
```
NuSkooler, line 2: Perhaps `set -e` ?

docker/bin/docker-entrypoint.sh
NuSkooler, line 4: nit: UPPER_CASE_VARS to match other scripts in the repo

docker/bin/docker-entrypoint.sh
```diff
@@ -0,0 +1,26 @@
...
+        if [ -n "$(find "$bbspath/$dir" -maxdepth 0 -type d -empty 2>/dev/null)" ]; then
+            cp -rpn $bbsstgp/$dir/* $bbspath/$dir/
+        else
+            echo "WARN skipped $bbspath/$dir - vol Not empty/not a new setup - possible bad state"
+        fi
+    done
+    ./oputil.js config new
+fi
+if [[ ! -f $bbspath/config/$configname ]]; then #make sure once more, otherwise pm2-runtime will loop if missing the config
+  echo "for some reason you have skipped configuration - enigma will not work. please run config"
```
NuSkooler, line 22: super nit: `ENiGMA 1/2`

docker/bin/docker-entrypoint.sh
```diff
@@ -0,0 +1,26 @@
...
+configname=config.hjson # this is the default name, this script is intended for easy get-go - make changes as needed
+
+# Setup happens when there is no existing config file
+if [[ ! -f $bbspath/config/$configname ]]; then
+    for dir in "${prepopvols[@]}"
+    do
+        if [ -n "$(find "$bbspath/$dir" -maxdepth 0 -type d -empty 2>/dev/null)" ]; then
+            cp -rpn $bbsstgp/$dir/* $bbspath/$dir/
+        else
+            echo "WARN skipped $bbspath/$dir - vol Not empty/not a new setup - possible bad state"
```
NuSkooler, line 16: nit: `WARN: Skipped $BBS_ROOT/$DIR: Volume not empty or not a new setup; Possible bad state"

...and maybe some blurb echoed out about actionable steps

docker/bin/docker-entrypoint.sh
abutbul: done in latest commit

docker/bin/docker-entrypoint.sh
abutbul: better safe than sorry!

docker/bin/docker-entrypoint.sh
abutbul: done in latest commit

docker/bin/docker-entrypoint.sh
abutbul: done in latest commit

abutbul:
@NuSkooler ready

NuSkooler:
@davestephens Any more thoughts on this? This seems good to me, but your eyes are probably better on this.

abutbul:
> @davestephens Any more thoughts on this? This seems good to me, but your eyes are probably better on this.

i'd really like some one else to try and deploy it as well. i'm currently the only tester afaik.

NuSkooler:
> > Thanks for the rundown, I like it --
> > Have you looked at the config's environment variable support? https://github.com/NuSkooler/enigma-bbs/blob/master/docs/configuration/config-files.md#environment-variables
> 
> missed answering that one, it's pretty neat. how do you see it being used in this flow?

I'm not 100% sure, actually :D There may be some ways to smooth env files -> config expansion though. Mostly just making you aware in case it spurred some thought from your side.

NuSkooler:
FYI https://github.com/NuSkooler/enigma-bbs/actions/runs/1766553445

docker/bin/docker-entrypoint.sh
```diff
@@ -0,0 +1,29 @@
...
+        fi
+    done
+    ./oputil.js config new
+fi
+if [[ ! -f $BBS_ROOT/config/$CONFIG_NAME ]]; then #make sure once more, otherwise pm2-runtime will loop if missing the config
+  printf "ERROR: Missing configuration - ENiGMA 1/2 will not work. please run config\n"
+  
+  exit 1
+else
+    pm2-runtime main.js
```
davestephens, line 28: Should `exec` here to replace the script with pm2 as PID 0.

docker/bin/docker-entrypoint.sh
abutbul: > FYI https://github.com/NuSkooler/enigma-bbs/actions/runs/1766553445

this fails in     tags: enigmabbs/enigma-bbs:latest

due to missing python3, my code should fix it

docker/bin/docker-entrypoint.sh
abutbul: I do not understand this request. what is the behavior you are trying to fix?
why is it so important to run as pm2 as PID0 if docker-entrypoint as  PID0 is handling its errors and passes it on?
will fix as soon as I understand the issue and the solution.

docker/bin/docker-entrypoint.sh
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env bash
+set -e
+
+# Set some vars
+PRE_POP_VOLS=("config" "mods" "art") # these are folders which contain runtime needed files, and need to be represented in the host
+BBS_ROOT=/enigma-bbs # install location
+BBS_STG_P=/enigma-bbs-pre # staging location for pre populated volumes (PRE_POP_VOLS)
```
davestephens, line 7: Maybe picky, bit I have a general distaste for opaque variable names. There's no limit in bash, so more descriptive and verbose variable names are preferable for maintainability.

docker/bin/docker-entrypoint.sh
abutbul: image
image

docker/bin/docker-entrypoint.sh
```diff
@@ -0,0 +1,29 @@
...
+
+# Setup happens when there is no existing config file
+if [[ ! -f $BBS_ROOT/config/$CONFIG_NAME ]]; then
+    for DIR in "${PRE_POP_VOLS[@]}"
+    do
+        if [ -n "$(find "$BBS_ROOT/$DIR" -maxdepth 0 -type d -empty 2>/dev/null)" ]; then
+            cp -rp $BBS_STG_P/$DIR/* $BBS_ROOT/$DIR/
+        else
+            printf "WARN: skipped $BBS_ROOT/$DIR: Volume not empty or not a new setup; Files required to run ENiGMA 1/2 may be missing.\n Possible bad state\n"
+            printf "INFO:you have mounted folders with existing data - but no existing config json.\n\nPossible solutions:\n1. Make sure all volumes are set correctly specifically config volume... \n2. Check your configuration name if non-default\n\n\n"
```
davestephens, line 18: `INFO: You....`

docker/bin/docker-entrypoint.sh
```diff
@@ -0,0 +1,29 @@
...
+        if [ -n "$(find "$BBS_ROOT/$DIR" -maxdepth 0 -type d -empty 2>/dev/null)" ]; then
+            cp -rp $BBS_STG_P/$DIR/* $BBS_ROOT/$DIR/
+        else
+            printf "WARN: skipped $BBS_ROOT/$DIR: Volume not empty or not a new setup; Files required to run ENiGMA 1/2 may be missing.\n Possible bad state\n"
+            printf "INFO:you have mounted folders with existing data - but no existing config json.\n\nPossible solutions:\n1. Make sure all volumes are set correctly specifically config volume... \n2. Check your configuration name if non-default\n\n\n" 
+        fi
+    done
+    ./oputil.js config new
+fi
+if [[ ! -f $BBS_ROOT/config/$CONFIG_NAME ]]; then #make sure once more, otherwise pm2-runtime will loop if missing the config
```
davestephens, line 23: `# Make sure...`

docker/bin/docker-entrypoint.sh
abutbul: can do but just to make sure i understand:
PRE_POP_VOLS=PRE_POPULATED_VOLUMES
BBS_ROOT=BBS_ROOT_DIR
BBS_STG_P=BBS_STAGING_PATH

docs/installation/docker.md
```diff
@@ -6,20 +6,43 @@ title: Docker
 for every operating system on the [Docker website](https://docs.docker.com/engine/install/).**
 
 ## Quick Start
+prepare a folder where you are going to save your bbs files.
+- Generate some config for your BBS: \
+you can perform this step from anywhere - but make sure to consistently run it from the same place to retain your config inside the docker guest
+```
+docker run -it -p 8888:8888 \
+--name "EnigmaBBS" \
```
davestephens, line 14: Worth using the correct casing for `ENIGMA` for the container name?

abutbul: yes!

abutbul:
@davestephens fixed all but "Should exec here to replace the script with pm2 as PID 0." please elaborate
@NuSkooler did you see my comment? you've built the old image, i know it fails this is why I started hacking at this. try and run my PR merged change.

NuSkooler:
> @NuSkooler did you see my comment? you've built the old image, i know it fails this is why I started hacking at this. try and run my PR merged change.

Yep makes sense

davestephens:
> @davestephens fixed all but "Should exec here to replace the script with pm2 as PID 0." please elaborate 

Bash doesn't pass signals to child processes, it just kills them. Therefore, pm2, and thus enigma, doesn't receive a sigkill and exit gracefully when someone issues a `docker stop`, it just gets nuked. If it's midway through a database or mail transaction for example, this is doubly important. 

Exec-ing pm2 is the correct approach to fix this, as the bash script is replaced by pm2 once the bash script completes (and thus receives signals), rather than leaving a weird process tree lying around. 

All you need to do is prefix pm2-runtime with exec.

abutbul:
> > @davestephens fixed all but "Should exec here to replace the script with pm2 as PID 0." please elaborate
> 
> Bash doesn't pass signals to child processes, it just kills them. Therefore, pm2, and thus enigma, doesn't receive a sigkill and exit gracefully when someone issues a `docker stop`, it just gets nuked. If it's midway through a database or mail transaction for example, this is doubly important.
> 
> Exec-ing pm2 is the correct approach to fix this, as the bash script is replaced by pm2 once the bash script completes (and thus receives signals), rather than leaving a weird process tree lying around.
> 
> All you need to do is prefix pm2-runtime with exec.

fixed
thank you for sharing your knowledge about this!

I learn fast, but need slow explaining

NuSkooler:
@abutbul for some reason I don't see the fix applied for `exec`. Did it get pushed? I could have sworn GitHub gave me a notification, but it's not showing in the diffs for me.

abutbul:
> @abutbul for some reason I don't see the fix applied for `exec`. Did it get pushed? I could have sworn GitHub gave me a notification, but it's not showing in the diffs for me.
it's weird, it's commited to my master https://github.com/abutbul/enigma-bbs/commit/7bbb2104b9644097a06959d32eb8775fc0981cb7 which is the branch in this PR

I will commit some empty space to see if this can get unstuck

abutbul:
weird but fixed it...

NuSkooler:
Thank you again!

PR #389, opened 2022-01-28
General bug fixes

cognitivegears:
This is just a few small bugfixes I found while creating documentation for the rest of the views (which is in progress, should have in the next day or two). Sorry for the grab-bag of changes, I'm just trying to find/fix small things as I go through each view type and the code. The following are included:

* In button_view.js the code didn't check for whether key was defined before using it to look up the mapping. I ran into some weird instability with this, where in some cases if the focus was on a button and I was entering hotkeys (switching windows etc) it would sometimes cause Enigma to die with an error about key not being defined. It must have been mistypes sending weird characters or something, but in any case anything that crashes is probably bad, so I put some defensive code around it that was used in other places in the code.
* In fse.js I changed the logic slightly so a from address doesn't have to be on the screen for the module to work (not needed, for example, on the new user feedback)
* text_view.js was padding one character further than specified in the config files.
* In view.js it was possible to setWidth wider than the space left available, because it wasn't taking into account the position.
* Removed some commented code that looked out of date.
* Fixed case of mask_edit_text_view.js where the first character is a mask character, but the second isn't - i.e. `# feet ## inches` would throw an assertion if you tried to backspace to the beginning of the input.

### DIFF
--- a/core/button_view.js
+++ b/core/button_view.js
@@ -21,8 +21,7 @@ function ButtonView(options) {
 util.inherits(ButtonView, TextView);
 
 ButtonView.prototype.onKeyPress = function(ch, key) {
-    let actionForKeyName = key ? key.name : ch;
-    if(this.isKeyMapped('accept', actionForKeyName) || ' ' === ch) {
+    if(this.isKeyMapped('accept', (key ? key.name : ch)) || ' ' === ch) {
         this.submitData = 'accept';
         this.emit('action', 'accept');
         delete this.submitData;
--- a/core/button_view.js
+++ b/core/button_view.js
@@ -21,7 +21,8 @@ function ButtonView(options) {
 util.inherits(ButtonView, TextView);
 
 ButtonView.prototype.onKeyPress = function(ch, key) {
-    if(this.isKeyMapped('accept', key.name) || ' ' === ch) {
+    let actionForKeyName = key ? key.name : ch;
+    if(this.isKeyMapped('accept', actionForKeyName) || ' ' === ch) {
         this.submitData = 'accept';
         this.emit('action', 'accept');
         delete this.submitData;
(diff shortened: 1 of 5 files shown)


core/button_view.js
```diff
@@ -21,24 +21,15 @@ function ButtonView(options) {
 util.inherits(ButtonView, TextView);
 
 ButtonView.prototype.onKeyPress = function(ch, key) {
-    if(this.isKeyMapped('accept', key.name) || ' ' === ch) {
+    let actionForKeyName = key ? key.name : ch;
```
NuSkooler, line 24: super nit: `const`

core/fse.js
```diff
@@ -791,10 +791,10 @@ exports.FullScreenEditorModule = exports.getModule = class FullScreenEditorModul
                     });
                 },
                 function prepareViewStates(callback) {
-                    var header = self.viewControllers.header;
-                    var from = header.getView(MciViewIds.header.from);
-                    from.acceptsFocus = false;
-                    //from.setText(self.client.user.username);
+                    let from = self.viewControllers.header.getView(MciViewIds.header.from);
+                    if (from !== undefined) {
```
NuSkooler, line 795: Probably just want `if (from)`

core/fse.js
```diff
@@ -825,10 +825,13 @@ exports.FullScreenEditorModule = exports.getModule = class FullScreenEditorModul
                             {
                                 const fromView = self.viewControllers.header.getView(MciViewIds.header.from);
                                 const area = getMessageAreaByTag(self.messageAreaTag);
-                                if(area && area.realNames) {
-                                    fromView.setText(self.client.user.properties[UserProps.RealName] || self.client.user.username);
-                                } else {
-                                    fromView.setText(self.client.user.username);
+
+				if(fromView !== undefined) {
```
NuSkooler, line 829: This isn't your doing, but just so you're aware there are newer APIs available that may help with some of this work in `menu_module.js` (available to all derived menus), and things like `client.user.getProperty()` to clean up some code as well.

NuSkooler:
This looks great -- ready to merge if you are.

core/button_view.js
cognitivegears: Fixed - since it was only used once went one further and just got rid of the unnecessary variable altogether.

core/fse.js
cognitivegears: Done.

core/fse.js
cognitivegears: Views first, then modules. :D I'm still learning, hoping to do some more around this when I tackle documenting the modules. This one was just for my own board (which is not even anywhere close to live yet, I'm having too much fun working on the code to actually spend time getting it going yet :) ) One of the nice things about working on documentation though is besides learning a lot about how it works, I also find little stuff like this, and sometimes it's easier to fix it than document it. lol

cognitivegears:
Should be good to go if you are ok with it.

PR #390, opened 2022-01-29
Feature/view documentation

cognitivegears:
As promised, here is the view documentation for the rest of the elements. I tried to go through the code, the existing documentation, and implement each one to see what options were valid, how they operated, etc. Please review, as I'm sure I probably missed something here and there.

Also, I did not create documentation for KeyValue view, as I'm not exactly sure what it is supposed to do. It acts like a hotkey like the documentation says, but despite having a "keys" option that sets which keys are allowed, the action is submitted immediately on key press no matter which key is pressed. I think this might not be the right behavior, but I decided to hold off rather than give (maybe bad) info.

Also, I did not add documentation for XY, I see what it is doing with positioning, but I'm not sure why it would be used. I decided to hold off until I understood more about it.

If possible can you please review these and let me know if there is anything I missed, etc? Thanks!


NuSkooler:
@cognitivegears Thank you! This is awesome work, I really appreciate it! I'll start going through these.

RE: `KeyView` -- I'll need to look at that. Doesn't sound right, but I know people are using it... might just be flying under the radar. 

`XY` is used for recording positions, generally to be referenced in code. For example, placement for other controls rendered dynamically at runtime.

PR #393, opened 2022-01-29
stop showing 'MRC_Watchdog@MRC_Watchdog_Process just left the server'…

christiansacks:
… in message window every time MRC server sends a watchdog

image

### DIFF
--- a/core/mrc.js
+++ b/core/mrc.js
@@ -325,8 +325,10 @@ exports.getModule = class mrcModule extends MenuModule {
                     }
 
                     default:
-                        this.addMessageToChatLog(message.body);
-                        break;
+                        if(!message.body.includes("MRC_Watchdog")) {
+                            this.addMessageToChatLog(message.body);
+                            break;
+                        }
                 }
 
             } else {


christiansacks:
I added `/me` also to mrc

core/mrc.js
```diff
@@ -430,6 +432,12 @@ exports.getModule = class mrcModule extends MenuModule {
                 this.processOutgoingMessage(newmsg, cmd[1]);
                 break;
 
+            case 'me':
+                const memsg = cmd.slice(1).join(' ');
+                const handle = this.state.alias;
+                this.sendMessageToMultiplexer("", "", this.state.room, "|15* |13"+handle+" "+stripMciColorCodes(memsg));
```
NuSkooler, line 438: You can use format strings, e.g. 
```
`|15* |13{handle} {stripMciColorCodes(memsg))`
```

christiansacks: interesting, I must have done something wrong as I tried that and the output just showed ".... {handle} {stripMciColorCodes(memsg)}" lol

Will have a play

christiansacks: ok I figured out how to use those, dunno if that's how you meant but it works still =)

stack-fault:
This will ne longer be required since the watchdog will not be broadcasted anymore

PR #396, opened 2022-02-02
Improve documentation - mobile support

cognitivegears:
This PR is to enable mobile support for the documentation. Currently, the documentation isn't really usable on mobile, as the sidebar takes up most of the screen. This change forces the menu to the bottom of the page on small screens and adds a menu button to jump to it. I've tested it out with Firefox responsive mode as well as Chrome. In addition, I went ahead and added a favicon and cleaned up just a few other things Lighthouse commented on. Finally, I updated the version of Jekyll so that it runs better locally, and added the relative links plugin so that it behaves the same way it does running on Github.


NuSkooler:
You're a beast, thank you!

PR #404, opened 2022-02-05
Change documentation to use collections + new features

cognitivegears:
Since I'm on a run of making Pull Requests for things that nobody actually asked for, but that I think are pretty cool :)
This PR makes a fairly heavy change to the documentation. It creates a Jekyll collection out of the documentation pages, introduces a GH Action to build the Jekyll site (needed since the really old version on GH didn't support all the functionality this needs) and reworks the navigation. The benefits are:

* No more having to update the nav.md (now nav.html) with each change. The navigation is created automatically! Instead just specify the order of pages within the _config.yml
* Next and previous links on each of the documentation pages. Particularly helpful on mobile, but works well on desktop too if you are reading whole sections of the manual.
* Fixed a bunch of linking, so links should now work both on github code view and the documentation site.
* Minified the resulting pages for moar speed.
* A few "missing" pages that existed but were not linked added to the menu.
* Much more recent version of Jekyll, liquid, bundler etc to squash bugs and add features.
* Can add additional plugins, even ones that are not on gh-pages.  (I have a cool idea for this for a future PR)

High level list of changes:
1) New github Action "jekyll.yml" to execute the build on check-ins to main. This also means I can run this on my fork so I can actually test things before submitting them :)
2) Add new documentation files to _config.yml. This is only for controlling the ordering, if you don't they'll just end up at the end of the list instead of being lost.
3) If you have a new category, add it to _data/sections.yml. This is just a lookup from directory names to titles for the navigation.
4) Documentation moved under a _docs collection directory. Set it up so they still publish to the same place though in case anyone had a bookmark, etc.

If you want to see what the results look like, check out my fork - https://cognitivegears.github.io/enigma-bbs/

Please let me know if you have any questions. Thanks!


NuSkooler:
> Since I'm on a run of making Pull Requests for things that nobody actually asked for, but that I think are pretty cool :)

I agree, these PRs are awesome! This looks really great, thanks!

Issue #222, opened 2018-12-10
Track client cursor position internally


NuSkooler:
**Issue**:
Currently the system relies on ANSI DSR (Device Status Report)/CPR sequences in many areas to discover / track client cursor positions. Not limited to:
- MCI position
- Pause
- Some work in the FSE

This has some major drawbacks:
- Does not work in all terminals. Some terminals such as mTCP's telnet only _partially_ work as their queue of DSR's is very small. Many older DOS terminals do not implement it at all.
- Introduces a lag (essentially, we must SYN/ACK the positions).
- May be wrong by the time we get the data, i.e.: if data scrolls (See #5)

Since we know the term dimensions, we should track the positions ourselves. This has some implications around the current implementation of Views, `ansi_term.js`'s goto(), etc.

Also related: Some code saves/restores (SCP – Save Cursor Position / RCP – Restore Cursor Position) which may need to go as well.

NuSkooler:
Big shout to @cognitivegears for the work on this!

PR #407, opened 2022-03-26
Removing Cursor Position Report

cognitivegears:
This PR is a potential fix for #222 track cursor position internally. This removes most usages of the cursor position report. The only remaining use is with the UTF8 detection - I added a configuration option for whether to use it, and even if it is used it should just time out instead of causing incompatibilities.

I'm hoping that you or someone else can try this one out as well. This works on my (internal only still) board, but I'd love some additional feedback on this one as well if possible.

NOTE - Although I did try to address some of the concerns of some of the other issues around art longer than the term height, I haven't been able to fully test this yet. Once we get the cursor position reports taken care of I'll circle back around and do some additional testing against those issues as well.

### DIFF
--- a/core/ansi_escape_parser.js
+++ b/core/ansi_escape_parser.js
@@ -21,9 +21,15 @@ function ANSIEscapeParser(options) {
     events.EventEmitter.call(this);
 
     this.column             = 1;
-    this.row                = 1;
     this.scrollBack         = 0;
     this.graphicRendition   = {};
+
+    if(options.startRow != null) {
+        this.row = options.startRow;
+    }
+    else {
+        this.row = 1;
+    }
 
     this.parseState = {
         re  : /(?:\x1b\x5b)([?=;0-9]*?)([ABCDHJKfhlmnpsu])/g,   //  eslint-disable-line no-control-regex
--- a/core/ansi_escape_parser.js
+++ b/core/ansi_escape_parser.js
@@ -190,10 +190,11 @@ function ANSIEscapeParser(options) {
 
 
                 self.emit('mci', {
-                    mci     : mciCode,
-                    id      : id ? parseInt(id, 10) : null,
-                    args    : args,
-                    SGR     : ansi.getSGRFromGraphicRendition(self.graphicRendition, true)
+                    position : [self.row, self.column],
+                    mci      : mciCode,
+                    id       : id ? parseInt(id, 10) : null,
+                    args     : args,
+                    SGR      : ansi.getSGRFromGraphicRendition(self.graphicRendition, true)
                 });
 
                 if(self.mciReplaceChar.length > 0) {
--- a/core/ansi_escape_parser.js
+++ b/core/ansi_escape_parser.js
@@ -21,15 +21,7 @@ function ANSIEscapeParser(options) {
     events.EventEmitter.call(this);
 
     this.column             = 1;
-    this.scrollBack         = 0;
     this.graphicRendition   = {};
-
-    if(!_.isNil(options?.startRow)) {
-        this.row = options.startRow;
-    }
-    else {
-        this.row = 1;
-    }
 
     this.parseState = {
         re  : /(?:\x1b\x5b)([?=;0-9]*?)([ABCDHJKfhlmnpsu])/g,   //  eslint-disable-line no-control-regex
@@ -42,10 +34,14 @@ function ANSIEscapeParser(options) {
         trailingLF          : 'default',    //  default|omit|no|yes, ...
     });
 
+
     this.mciReplaceChar     = miscUtil.valueWithDefault(options.mciReplaceChar, '');
     this.termHeight         = miscUtil.valueWithDefault(options.termHeight, 25);
     this.termWidth          = miscUtil.valueWithDefault(options.termWidth, 80);
     this.trailingLF         = miscUtil.valueWithDefault(options.trailingLF, 'default');
+
+
+    this.row = Math.min(options?.startRow ?? 1, this.termHeight);
 
     self.moveCursor = function(cols, rows) {
         self.column += cols;
@@ -75,14 +71,11 @@ function ANSIEscapeParser(options) {
     };
 
     self.clearScreen = function() {
-        //  :TODO: should be doing something with row/column?
+        self.column = 1;
+        self.row    = 1;
         self.emit('clear screen');
     };
 
-    /*
-    self.rowUpdated = function() {
-        self.emit('row update', self.row + self.scrollBack);
-    };*/
 
     self.positionUpdated = function() {
         self.emit('position update', self.row, self.column);
--- a/core/ansi_escape_parser.js
+++ b/core/ansi_escape_parser.js
@@ -215,6 +215,9 @@ function ANSIEscapeParser(options) {
     }
 
     self.reset = function(input) {
+        self.column = 1;
+        self.row = Math.min(options?.startRow ?? 1, self.termHeight);
+
         self.parseState = {
             //  ignore anything past EOF marker, if any
             buffer  : input.split(String.fromCharCode(0x1a), 1)[0],
--- a/core/ansi_escape_parser.js
+++ b/core/ansi_escape_parser.js
@@ -24,7 +24,7 @@ function ANSIEscapeParser(options) {
     this.scrollBack         = 0;
     this.graphicRendition   = {};
 
-    if(options.startRow != null) {
+    if(!_.isNil(options.startRow)) {
         this.row = options.startRow;
     }
     else {
--- a/core/ansi_escape_parser.js
+++ b/core/ansi_escape_parser.js
@@ -24,7 +24,7 @@ function ANSIEscapeParser(options) {
     this.scrollBack         = 0;
     this.graphicRendition   = {};
 
-    if(!_.isNil(options.startRow)) {
+    if(!_.isNil(options?.startRow)) {
         this.row = options.startRow;
     }
     else {
--- a/core/config_default.js
+++ b/core/config_default.js
@@ -23,6 +23,12 @@ module.exports = () => {
             // during the connect process, but provides better autoconfiguration of utf-8
             checkUtf8Encoding : true,
 
+            // Checking the ANSI home position also requires the use of cursor position reports, which are not
+            // supported on all terminals. Using this with a terminal that does not support cursor position reports
+            // results in a 3 second delay during the connect process, but works around positioning problems with
+            // non-standard terminals.
+            checkAnsiHomePosition: true,
+
             // List of terms that should be assumed to use cp437 encoding
             cp437TermList     : ['ansi', 'pcansi', 'pc-ansi', 'ansi-bbs', 'qansi', 'scoansi', 'syncterm', 'ansi-256color', 'ansi-256color-rgb'],
             // List of terms that should be assumed to use utf8 encoding
--- a/core/config_default.js
+++ b/core/config_default.js
@@ -17,6 +17,18 @@ module.exports = () => {
             achievementFile : 'achievements.hjson',
         },
 
+        term : {
+            // checkUtf8Encoding requires the use of cursor position reports, which are not supported on all terminals.
+            // Using this with a terminal that does not support cursor position reports results in a 2 second delay
+            // during the connect process, but provides better autoconfiguration of utf-8
+            checkUtf8Encoding : true,
+
+            // List of terms that should be assumed to use cp437 encoding
+            cp437TermList     : ['ansi', 'pcansi', 'pc-ansi', 'ansi-bbs', 'qansi', 'scoansi', 'syncterm', 'ansi-256color', 'ansi-256color-rgb'],
+            // List of terms that should be assumed to use utf8 encoding
+            utf8TermList      : ['xterm', 'linux', 'screen', 'dumb', 'rxvt', 'konsole', 'gnome', 'x11 terminal emulator'],
+        },
+
         users : {
             usernameMin         : 2,
             usernameMax         : 16,   //  Note that FidoNet wants 36 max
--- a/core/connect.js
+++ b/core/connect.js
@@ -12,6 +12,58 @@ const async     = require('async');
 
 exports.connectEntry    = connectEntry;
 
+function ansiDiscoverHomePosition(client, cb) {
+    //
+    //  We want to find the home position. ANSI-BBS and most terminals
+    //  utilize 1,1 as home. However, some terminals such as ConnectBot
+    //  think of home as 0,0. If this is the case, we need to offset
+    //  our positioning to accommodate for such.
+    //
+
+
+    if( !Config().term.checkAnsiHomePosition ) {
+        // Skip (and assume 1,1) if the home position check is disabled.
+        return cb(null);
+    }
+
+    const done = (err) => {
+        client.removeListener('cursor position report', cprListener);
+        clearTimeout(giveUpTimer);
+        return cb(err);
+    };
+
+    const cprListener = function(pos) {
+        const h = pos[0];
+        const w = pos[1];
+
+        //
+        //  We expect either 0,0, or 1,1. Anything else will be filed as bad data
+        //
+        if(h > 1 || w > 1) {
+            client.log.warn( { height : h, width : w }, 'Ignoring ANSI home position CPR due to unexpected values');
+            return done(Errors.UnexpectedState('Home position CPR expected to be 0,0, or 1,1'));
+        }
+
+        if(0 === h & 0 === w) {
+            //
+            //  Store a CPR offset in the client. All CPR's from this point on will offset by this amount
+            //
+            client.log.info('Setting CPR offset to 1');
+            client.cprOffset = 1;
+        }
+
+        return done(null);
+    };
+
+    client.once('cursor position report', cprListener);
+
+    const giveUpTimer = setTimeout( () => {
+        return done(Errors.General('Giving up on home position CPR'));
+    }, 3000);   //  3s
+
+    client.term.write(`${ansi.goHome()}${ansi.queryPos()}`);    //  go home, query pos
+}
+
 function ansiAttemptDetectUTF8(client, cb) {
     //
     //  Trick to attempt and detect UTF-8. While there is a lot more than
@@ -172,6 +224,11 @@ function connectEntry(client, nextMenu) {
             function basicPrepWork(callback) {
                 term.rawWrite(ansi.queryDeviceAttributes(0));
                 return callback(null);
+            },
+            function discoverHomePosition(callback) {
+                ansiDiscoverHomePosition(client, () => {
+                    return callback(null);  //  we try to continue anyway
+                });
             },
             function queryTermSizeByNonStandardAnsi(callback) {
                 ansiQueryTermSizeIfNeeded(client, err => {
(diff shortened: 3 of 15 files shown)


NuSkooler:
Thanks!

I've just skimmed over this a tad so far, but this looks great! I'll pull down and start playing with it. I have a lot of notes to go through around this mess actually -- it's something I should have done long ago :laughing:

core/menu_module.js
```diff
@@ -138,9 +130,14 @@ exports.MenuModule = class MenuModule extends PluginModule {
                         return callback(null);
```
NuSkooler, line 130:
```
diff --git a/core/menu_module.js b/core/menu_module.js
index ef6ad831..86ff119f 100644
--- a/core/menu_module.js
+++ b/core/menu_module.js
@@ -127,7 +127,7 @@ exports.MenuModule = class MenuModule extends PluginModule {
                 },
                 function displayPauseIfRequested(callback) {
                     if(!self.shouldPause()) {
-                        return callback(null);
+                        return callback(null, null);
                     }
 
                     if(self.client.term.termHeight > 0 && pausePosition.row > self.client.termHeight) {
```

core/config_default.js
```diff
@@ -17,6 +17,18 @@ module.exports = () => {
             achievementFile : 'achievements.hjson',
         },
 
+        term : {
+            // checkUtf8Encoding requires the use of cursor position reports, which are not supported on all terminals.
```
NuSkooler, line 21: Nice!

core/ansi_escape_parser.js
```diff
@@ -21,10 +21,16 @@ function ANSIEscapeParser(options) {
     events.EventEmitter.call(this);
 
     this.column             = 1;
-    this.row                = 1;
     this.scrollBack         = 0;
     this.graphicRendition   = {};
 
+    if(options.startRow != null) {
```
NuSkooler, line 27: Careful with `null` here due to JS `undefined`

core/connect.js
```diff
@@ -4,58 +4,14 @@
...
 exports.connectEntry    = connectEntry;
 
-function ansiDiscoverHomePosition(client, cb) {
-    //
-    //  We want to find the home position. ANSI-BBS and most terminals
-    //  utilize 1,1 as home. However, some terminals such as ConnectBot
-    //  think of home as 0,0. If this is the case, we need to offset
-    //  our positioning to accommodate for such.
-    //
-    const done = (err) => {
```
NuSkooler, line 21: I think we might still want this. *If* we can query and get a response, this helps hack around 1,1 (standard) vs 0,0 (non-standard) home positions. If we can't detect it we should time out and no harm no foul.

core/menu_module.js
```diff
@@ -106,30 +103,25 @@ exports.MenuModule = class MenuModule extends PluginModule {
                         return callback(Errors.MissingConfig('Prompt specified but no "promptConfig" block found'));
                     }
 
+                    const options = self.menuConfig.config;
+
+                    if(artData != null && artData.height != null) {
```
NuSkooler, line 108: You should be able to do something like:
`if (_.isNumber(artData?.height)) { ... }`

NuSkooler:
Looks like there is a placement issue some MCI's in menus and/or prompts. SyncTERM, NetRunner, and MagiTerm all look something like this:
image

Will follow after poking at it a bit if I get some time in a few.

core/menu_module.js
```diff
@@ -512,7 +509,7 @@ exports.MenuModule = class MenuModule extends PluginModule {
 
         this.optionalMoveToPosition(position);
 
-        return theme.displayThemedPause(this.client, cb);
```
NuSkooler, line 515: 2nd param is a object of `options`, so I think you're wanting this:
```
return theme.displayThemedPause(this.client, { position }, cb)
```

cognitivegears:
> Looks like there is a placement issue some MCI's in menus and/or prompts. SyncTERM, NetRunner, and MagiTerm all look something like this:
(cut)
> Looks like `self.row` ends up being very large, then the terms just cut it at max. I don't have a ton of time to look at this tonight, but will follow up if I find something.

Thanks - my main menu has changed so I don't see this - I'll try reverting back to the standard theme and see if I can duplicate this. Most of the cases I saw where the mci location got off was if there was multiple assets show  but it wasn't tracking between them - but that doesn't sound like the case here if self.row is large - maybe the opposite case where there is a screen clear or positioning that isn't being taken into account - not sure but I'll try to reproduce this and see if I can find out more too, thanks.

core/menu_module.js
cognitivegears: Sorry more of my old school js slipping through 🙂 thanks I'll update this

core/connect.js
cognitivegears: Good point. I'll put this back in, but maybe also put that behind a config option if you agree? That way someone who really doesn't want them can still avoid it (at the risk of some broken clients) but by default it can still do the check.

We could also theoretically combine this test with the UTF8 check (since it already does 2 stages) if it wasn't skipping that one for the "nix" clients to make only one place it does a position check still. Thoughts?

core/ansi_escape_parser.js
cognitivegears: Thanks I'll change this.

core/menu_module.js
cognitivegears: Oops sorry I'll change this, thanks!

core/connect.js
NuSkooler: I'm fine with putting it behind a configuration option, keeping default behavior -- that sounds great.

The problem with the UTF-8 check (kludge!) is it can bork up things with "BBS" clients. Aside: I really need to get around to cleaning up the names of all of this stuff too...

NuSkooler:
This is working pretty good for me under SyncTERM and NetRunner. MagiTerm still has the issue described prior. I need to go through quite a few other terminals as well.

```
diff --git a/core/ansi_escape_parser.js b/core/ansi_escape_parser.js
index c6d1276c..e9a6af91 100644
--- a/core/ansi_escape_parser.js
+++ b/core/ansi_escape_parser.js
@@ -21,16 +21,8 @@ function ANSIEscapeParser(options) {
     events.EventEmitter.call(this);
 
     this.column             = 1;
-    this.scrollBack         = 0;
     this.graphicRendition   = {};
 
-    if(options.startRow != null) {
-        this.row = options.startRow;
-    }
-    else {
-        this.row = 1;
-    }
-
     this.parseState = {
         re  : /(?:\x1b\x5b)([?=;0-9]*?)([ABCDHJKfhlmnpsu])/g,   //  eslint-disable-line no-control-regex
     };
@@ -47,6 +39,8 @@ function ANSIEscapeParser(options) {
     this.termWidth          = miscUtil.valueWithDefault(options.termWidth, 80);
     this.trailingLF         = miscUtil.valueWithDefault(options.trailingLF, 'default');
 
+    this.row = Math.min(options.startRow || 1, this.termHeight)
+
     self.moveCursor = function(cols, rows) {
         self.column += cols;
         self.row    += rows;
@@ -71,19 +65,14 @@ function ANSIEscapeParser(options) {
         delete self.savedPosition;
 
         self.positionUpdated();
-        //      self.rowUpdated();
     };
 
     self.clearScreen = function() {
-        //  :TODO: should be doing something with row/column?
+        self.column = 1;
+        self.row = 1;
         self.emit('clear screen');
     };
 
-    /*
-    self.rowUpdated = function() {
-        self.emit('row update', self.row + self.scrollBack);
-    };*/
-
     self.positionUpdated = function() {
         self.emit('position update', self.row, self.column);
     };
diff --git a/core/menu_module.js b/core/menu_module.js
index ef6ad831..518cc34f 100644
--- a/core/menu_module.js
+++ b/core/menu_module.js
@@ -127,7 +127,7 @@ exports.MenuModule = class MenuModule extends PluginModule {
                 },
                 function displayPauseIfRequested(callback) {
                     if(!self.shouldPause()) {
-                        return callback(null);
+                        return callback(null, null);
                     }
 
                     if(self.client.term.termHeight > 0 && pausePosition.row > self.client.termHeight) {
@@ -509,7 +509,7 @@ exports.MenuModule = class MenuModule extends PluginModule {
 
         this.optionalMoveToPosition(position);
 
-        return theme.displayThemedPause(this.client, position, cb);
+        return theme.displayThemedPause(this.client, { position }, cb);
     }
 
     promptForInput( { formName, formId, promptName, prevFormName, position } = {}, options, cb) {
```

core/menu_module.js
cognitivegears: Actually - as it turns out conditional chaining would require an upgrade to the `package.json` to upgrade eslint, and adding `"ecmaVersion": 2020` to the `.eslintrc.json`

I can do that, but it would also up the minimum version of Node.js as Node 12 doesn't support conditional chaining (looks like the minimum version for this is Node 14.) Are you ok with upping the minimum version?

core/menu_module.js
cognitivegears: Fixed.

core/ansi_escape_parser.js
cognitivegears: Fixed.

(I believe in the original it should have already been checking for both null and undefined since null == undefined, but it is a good call this makes it more explicit.

core/connect.js
cognitivegears: Added it back in with a configuration option, `term.checkAnsiHomePosition`

core/menu_module.js
NuSkooler: Yep let's do it!

cognitivegears: I went ahead and added the more verbose option for now. Let me know if you want me to bump the Node version instead.

core/menu_module.js
cognitivegears: I think you are right, and I believe that was the cause of the issue above as well. This was used in `theme.js` as options.row, which is why this worked even with it set to `position`. However, above in the `options.clearScreen` section, it was adding it as `options.position.row`. I fixed it to use `options.position.row` in all locations, and changed the `position` to `options.position`. Hopefully that fixes the large self.row value above.

core/menu_module.js
cognitivegears: Ok I'll upgrade the minimum version as well to Node 14

cognitivegears: I bumped up the minimum version of Node to 14, and changed the ECMAScript version to 2020.

I also made an assumption and went ahead and bumped the enigma-bbs version to 0.0.14-beta. I figured a new Node version plus removing CPR probably deserves a version bump if you agree?

core/ansi_escape_parser.js
```diff
@@ -21,10 +21,16 @@ function ANSIEscapeParser(options) {
     events.EventEmitter.call(this);
 
     this.column             = 1;
-    this.row                = 1;
     this.scrollBack         = 0;
     this.graphicRendition   = {};
 
+    if(!_.isNil(options?.startRow)) {
```
NuSkooler, line 27: I'm not sure if you saw it in the diff, but `this.row = Math.min(options.startRow || 1, this.termHeight)` after establishing termHeight ensures we don't end up with wacky values.

...but it should probably check `_.isNumber()` since startRow of 0 is actually acceptable for some wonky terminals.

core/theme.js
```diff
@@ -583,12 +585,15 @@ function displayThemedPrompt(name, client, options, cb) {
                     return callback(null, promptConfig, artInfo);
                 }
 
-                client.once('cursor position report', pos => {
-                    artInfo.startRow = pos[0] - artInfo.height;
-                    return callback(null, promptConfig, artInfo);
-                });
+                if(!_.isNil(options?.position?.row)) {
```
NuSkooler, line 588: `_.isNumber()`?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "enigma-bbs",
-  "version": "0.0.12-beta",
+  "version": "0.0.14-beta",
```
NuSkooler, line 3: This line is the enigma version. Though I think we should probably bump to `0.0.13` for this change.

core/menu_module.js
NuSkooler: Agree 100% that we should bump the enig version. We're at 0.0.12, so we can move to 0.0.13-beta now though.

core/menu_module.js
NuSkooler: I should have a bit of time later to pull down the latest and run through some tests. I need to round up some of the terminals to test against. Pretty excited about these changes!

core/ansi_escape_parser.js
cognitivegears: I'm so sorry - I was on mobile when I first looked at your comments, and totally missed this. Should be fixed now. Since we just updated to ES2020 anyway I just used the nullish coalescing operator (??) since that should do the right thing with 0 as well without the extra check. Happy to change it if you want though, lmk.

package.json
cognitivegears: Sorry, I was being silly, skipping .13 as an "unlucky number" lol. Changed it back to the actually sane value 0.0.13-beta now.

core/theme.js
cognitivegears: Fixed.

core/ansi_escape_parser.js
NuSkooler: No reason to apologize! I'm happy either way, but we do need to allow 0 and probably should cap at termHeight.

core/menu_module.js
NuSkooler: OK found the offset issue, see the diff below. The interesting part is where I'm now using `Object.assign()`. Previously, the menu config was mutated with `options.startRow = artData.height + 1;` -- instead, a copy is now made so it only applies for the particular instance.

```
diff --git a/core/ansi_escape_parser.js b/core/ansi_escape_parser.js
index 5417cef5..438f6203 100644
--- a/core/ansi_escape_parser.js
+++ b/core/ansi_escape_parser.js
@@ -215,6 +215,9 @@ function ANSIEscapeParser(options) {
     }
 
     self.reset = function(input) {
+        self.column = 1;
+        self.row = Math.min(options?.startRow ?? 1, self.termHeight);
+
         self.parseState = {
             //  ignore anything past EOF marker, if any
             buffer  : input.split(String.fromCharCode(0x1a), 1)[0],
diff --git a/core/menu_module.js b/core/menu_module.js
index 7669c2ee..bfddbd18 100644
--- a/core/menu_module.js
+++ b/core/menu_module.js
@@ -103,7 +103,7 @@ exports.MenuModule = class MenuModule extends PluginModule {
                         return callback(Errors.MissingConfig('Prompt specified but no "promptConfig" block found'));
                     }
 
-                    const options = self.menuConfig.config;
+                    const options = Object.assign({}, self.menuConfig.config);
 
                     if(_.isNumber(artData?.height)) {
                         options.startRow = artData.height + 1;
```

cognitivegears: Wow good catch. Done.

NuSkooler:
We'll need to get the `WHATSNEW.md` updated as well as the  `UPGRADE.md` (probably just a placeholder here, but it comes along with a version bump)

cognitivegears:
Sorry got stuck in some work stuff, should be able to check these in tomorrow. I saw your updates about the terminal testing - I'm worried about those status line issues. I'll dig in more first, but if it is that they are actually displaying less lines then they are reporting then I'm wondering if we could do something similar-ish to the 0,0 offset testing. We could try a single additional CPR on terminals at the beginning (with a timeout in case needed) where after going to the top line we then print 24 newlines (or whatever the reported term height - 1) and then do a CPR to see whether it is really term height - 1. Effectively (hopefully) getting the range for the window (0-24, 1-25, 1-24 etc) Thoughts?

NuSkooler:
>  I'm wondering if we could do something similar-ish to the 0,0 offset testing

I do think we could do something like this and get some of the more modern browsers or at least those that support CPR. A number of them are DOS terminals that simply don't have that support either. Perhaps worth a shot though.

In general, it's been a lot safer to rock 79x24 columns due to this non-standard stuff. A number of boards of course just ask you "What is the last line you see" or to simply just type in the number/etc. due to these issues.

BTW: I've been out of town as well. I'll try to get caught up on this and continue testing/poking at it. In general this is working very well. The status bar issue can be considered a separate effort I think.

cognitivegears:
> In general, it's been a lot safer to rock 79x24 columns due to this non-standard stuff.

Agreed - let me try adding something to the documentation to that effect.

> A number of boards of course just ask you "What is the last line you see" or to simply just type in the number/etc. due to these issues.
> The status bar issue can be considered a separate effort I think.

After this one is done, that was going to be my next PR I hope. Here's what I'm thinking:

Add a new encoding "ascii" with conversions that just strips anything past character code 127 (and probably some of the low values as well.) This wouldn't be used normally, but:

Add a configurations to `loginServers.telnet` and `loginServers.ssh`, something like `characterEncodingBehavior` with values `auto` (default) , `utf8`, `cp437`, `ascii` or `ask` (and other supported encodings if others are added later).

Similarly add `terminalGeometryBehavior` with `auto` (default), `fixed`, or `ask`. For `fixed` on this one there would be additional `.geometry.width` and `.geometry.height` settings (and maybe others).

The fixed settings would be useful if supporting multiple ports. i.e. "Connect to port 8887 for no formatting, port 8888 with retro BBS clients or port 8889 with modern UTF-8 terminals." or "Connect to port 4040 for 40 column or 8080 for 80 column." etc.

`ask` would just in-effect defer the configuration, skipping the current rules, and go to the `firstMenu`. All settings would be baseline, however - ASCII only, 40 column (but it doesn't really matter) etc. with the intention that these menus would be very basic ASCII screens to input the manual values, then set the properties appropriately (with a new module for this) before continuing to the login.

Finally, I'd like to add a `terminalAnsiBehavior` with `auto` (default), `ask`, `off`, `on`. This would require some additional work through the system to make positioning etc optional, but the more I look at it I don't think it's that bad - there are a lot of places, but in a lot of them similar strategies could be used. And if it only supported a subset of the MCI's etc that would probably be okay too.

The end result is a system that can operate exactly how it does now, but can also support a very, very large percentage of clients, even simple dumb terms and old 8-bit term programs without any ANSI or extended ASCII support. And, these options help with those weird edge cases that I can't figure out what to do with besides. For example, RIPterm identifies itself as `vt100`, but it uses `cp437`. I can't think of a good way of getting it to work without potentially breaking other clients. But if that was important to me as a sysop, I could open up a second port that allows users to configure the output manually for terminals like this.

Sorry - that was pretty long-winded. If you agree though, I'm hoping we can finish out the testing (I'm trying to get a few more terminal clients installed to help with that too) and close this one, then do that as additional work after?

NuSkooler:
> After this one is done, that was going to be my next PR I hope. Here's what I'm thinking

I really like a lot of the ideas you have here. I think we can achieve most of this perhaps via some expanded ACS (ie: connected port -- currently the servers only listen on single ports, but that could be adapted easily). This would keep the logic out of servers and more the spirit of enigma which is the menu system -- allowing ops to define it however they please. The configurations you describe could still be applied, but perhaps in the `connect.js` (which users can fully replace or skip as-is via their `menu.hjson`).

Hopefully that makes sense. Let's take this to a new feature ticket. This could solve a lot of issues with these old/probematic terminals, I agree.

NuSkooler:
This looks ready to go to me -- any objections?

cognitivegears:
Almost - I need to check in the `WHATSNEW.md` and `UPGRADE.md` - mostly done I was just waiting to see if there were any more changes. I will check those in tomorrow if that's ok.

cognitivegears:
> This looks ready to go to me -- any objections?

Checked in the UPGRADE.md and WHATSNEW.md - if those look good to you then I think we are ready to go.

PR #420, opened 2022-05-08
Update dockerfile to Node 14

georgzoeller:
Node 12 breaks, install.sh was updated but docker file was forgotten


Thraka:
I ran into the same thing. This fixed it for me.

NuSkooler:
Awesome, thank you!

PR #427, opened 2022-06-09
Updated ssh documentation

cognitivegears:
It looks like the previous suggestion of using OpenSSL no longer works. Using the OpenSSL method results in:

```
Error initializing: Error: Cannot parse privateKey: Unsupported key format
```

However, using ssh-keygen does appear to work well, at least under my testing. While I was at it I added a bit more documentation on the ssh module. Additional improvements are incoming.

### DIFF
--- a/docs/_docs/servers/loginservers/ssh.md
+++ b/docs/_docs/servers/loginservers/ssh.md
@@ -11,14 +11,18 @@ Entries available under `config.loginServers.ssh`:
 | Item | Required | Description |
 |------|----------|-------------|
 | `privateKeyPem` | :-1: | Path to private key file. If not set, defaults to `./config/ssh_private_key.pem` |
-| `privateKeyPass` | :+1: | Password to private key file.
+| `privateKeyPass` | :+1: | Password to private key file. *
 | `firstMenu` | :-1: | First menu an SSH connected user is presented with. Defaults to `sshConnected`. |
-| `firstMenuNewUser` | :-1: | Menu presented to user when logging in with one of the usernames found within `users.newUserNames` in your `config.hjson`. Examples include `new` and `apply`. |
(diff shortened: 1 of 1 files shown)


NuSkooler:
Thanks! 

Is this something that broke with a particular version of OpenSSL? Also, I'd prefer to use `--long` style options for readability, if you wouldn't mind.

cognitivegears:
I'm afraid I'm not sure on the version of OpenSSL - I presume it's more likely a change with the ssh2 library, but unfortunately I don't have the history on it, as I was just now setting up ssh for testing with Enigma.

As far as the options, unfortunately ssh-keygen doesn't look like it supports long style options. However, I definitely agree and so instead I also added a section below the command now describing the options and values to help with implementation.

NuSkooler:
Fair enough. Let's keep in the older examples as "Alternatives" that may work in other OpenSSL versions just in case. Both of the commands still work for me, e.g. with `OpenSSL 1.0.2k-fips  26 Jan 2017`

cognitivegears:
> Fair enough. Let's keep in the older examples as "Alternatives" that may work in other OpenSSL versions just in case. Both of the commands still work for me, e.g. with `OpenSSL 1.0.2k-fips 26 Jan 2017`

Thanks - updated. Note: I have `OpenSSL 3.0.2 15 Mar 2022 (Library: OpenSSL 3.0.2 15 Mar 2022) `. Looks like something happened between those two versions that may have broken compatibility. I have read that ssh-keygen uses OpenSSL (at least until very recently) under the covers, so there should be a modified command line that should work with even OpenSSL 3.0 but unfortunately I haven't had much luck finding what needs to be changed, and with the additional option of OpenSSH (ssh-keygen) I'm not inclined to look too much further into it unless you feel that it's important to do so.

PR #441, opened 2022-07-28
Feature/choose encoding

cognitivegears:
This is just a small change to allow for manually selecting encoding. By itself it's pretty simple, with a single method to allow a new encoding to be chosen from a menu. I'm planning on building on this functionality from here, but wanted to put in the base feature first.

### DIFF
--- a/core/system_menu_method.js
+++ b/core/system_menu_method.js
@@ -247,13 +247,6 @@ function setClientEncoding(callingMenu, formData, extraArgs, cb) {
     const client = callingMenu.client;
     let encoding = formData.value.encoding;
 
-    client.log.info(
-        { encoding: encoding, currentEncoding: client.term.outputEncoding },
-        'Setting client encoding.'
-    );
-
-    encoding = encoding || client.term.outputEncoding;
-
     client.term.outputEncoding = encoding;
 
     return callingMenu.nextMenu(cb);
--- a/core/system_menu_method.js
+++ b/core/system_menu_method.js
@@ -16,6 +16,7 @@ const iconv = require('iconv-lite');
 
 exports.login = login;
 exports.login2FA_OTP = login2FA_OTP;
+exports.setClientEncoding = setClientEncoding;
 exports.logoff = logoff;
 exports.prevMenu = prevMenu;
 exports.nextMenu = nextMenu;
@@ -239,6 +240,23 @@ function nextArea(callingMenu, formData, extraArgs, cb) {
             return reloadMenu(callingMenu, cb);
         }
     );
+}
+
+function setClientEncoding(callingMenu, formData, extraArgs, cb) {
+    // TODO: Also add other encoding types
+    const client = callingMenu.client;
+    let encoding = formData.value.encoding;
+
+    client.log.info(
+        { encoding: encoding, currentEncoding: client.term.outputEncoding },
+        'Setting client encoding.'
+    );
+
+    encoding = encoding || client.term.outputEncoding;
+
+    client.term.outputEncoding = encoding;
+
+    return callingMenu.nextMenu(cb);
 }
 
 function sendForgotPasswordEmail(callingMenu, formData, extraArgs, cb) {
--- a/docs/_docs/configuration/menu-hjson.md
+++ b/docs/_docs/configuration/menu-hjson.md
@@ -125,7 +125,7 @@ Many built in global/system methods exist. Below are a few. See [system_menu_met
 | `nextConf` | Sets the users message conference to the next available. |
 | `prevArea` | Sets the users message area to the previous available. |
 | `nextArea` | Sets the users message area to the next available. |
-| `setClientEncoding` | Sets the client encoding to either cp437 or utf-8. |
+| `setClientEncoding` | Sets the client encoding (such as cp437 and utf-8.) |
 
 ## Example
 Let's look a couple basic menu entries:
--- a/docs/_docs/configuration/menu-hjson.md
+++ b/docs/_docs/configuration/menu-hjson.md
@@ -76,7 +76,7 @@ Menus may also support more than one layout type by using a *MCI key*. A MCI key
 For more information on views and associated MCI codes, see [MCI Codes](../art/mci.md).
 
 ## Submit Handlers
-When a form is submitted, it's data is matched against a *submit handler*. When a match is found, it's *action* is performed.
+When a form is submitted, it's data is matched against a *submit handler*. When a match is found, it's *action* is performed. Note: Setting the value explicitly to null matches against any value.
 
 ### Submit Actions
 Submit actions are declared using the `action` member of a submit handler block. Actions can be kick off system/global or local-to-module methods, launch other menus, etc.
@@ -368,7 +368,7 @@ newUserApplicationPre: {
 
 Enigma½ tries to automatically determine the proper encoding for a client when it connects. Unfortunately, there are cases where the wrong encoding can be selected, resulting in terminal programs that are not supported. If your user base contains users that would like to connect with unsupported clients, one solution is to offer manual encoding selection.
 
-A new system method has been added to support this, @systemMethod:setClientEncoding.
+This can be accomplished with the system method `@systemMethod:setClientEncoding`.
 
 ### Simple example
 
@@ -385,7 +385,6 @@ telnetConnected: {
 clientSelectEncoding: {
     art: CLTSEL.ASC
     next: matrix
-    config: { interrupt: realtime }
     form: {
         0: {
             mci: {
@@ -410,11 +409,7 @@ clientSelectEncoding: {
             submit: {
               *: [
                 {
-                  value: { encoding: "utf-8" }
-                  action: @systemMethod:setClientEncoding
-                }
-                {
-                  value: { encoding: "cp437" }
+                  value: { encoding: null }
                   action: @systemMethod:setClientEncoding
                 }
               ]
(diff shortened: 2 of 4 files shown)


NuSkooler:
@cognitivegears Sweet! I'll start adding some comments. I think a number of things may be able to be simplified, but I like it! 

With just the one system method alone, one could make a `choose_encoding.js` mod for example that would toggle between UTF-8 and CP438 (/ others) visually like some of the boards do. Lots of possibilities.

core/system_menu_method.js
```diff
@@ -241,6 +242,23 @@ function nextArea(callingMenu, formData, extraArgs, cb) {
     );
 }
 
+function setClientEncoding(callingMenu, formData, extraArgs, cb) {
+    // TODO: Also add other encoding types
```
NuSkooler, line 246: In `client_term.js` the encoding property is checked, but it's just ignored and logged if a unsupported encoding is passed. Perhaps we should check it here, e.g.

```
iconv.encodingExists(...)
```

core/system_menu_method.js
```diff
@@ -241,6 +242,23 @@ function nextArea(callingMenu, formData, extraArgs, cb) {
...
+    const client = callingMenu.client;
+    let encoding = formData.value.encoding;
+
+    client.log.info(
```
NuSkooler, line 250: This log could probably be moved to `client_term.js` as well since that would capture all paths.

docs/_docs/configuration/menu-hjson.md
```diff
@@ -125,6 +125,7 @@ Many built in global/system methods exist. Below are a few. See [system_menu_met
 | `nextConf` | Sets the users message conference to the next available. |
 | `prevArea` | Sets the users message area to the previous available. |
 | `nextArea` | Sets the users message area to the next available. |
+| `setClientEncoding` | Sets the client encoding to either cp437 or utf-8. |
```
NuSkooler, line 128: Suggest instead of "either...", "such as..." or similar as other encoding are technically supported and hopefully some others down the line.

docs/_docs/configuration/menu-hjson.md
```diff
@@ -362,3 +363,188 @@ newUserApplicationPre: {
...
+                          text: C) CP437
+                          data: cp437
+                      }
+                    ]
+                }
+            }
+            submit: {
+              *: [
+                {
+                  value: { encoding: "utf-8" }
```
NuSkooler, line 413: If you wanted, you could just have a single handler and a value of null (match any)

docs/_docs/configuration/menu-hjson.md
```diff
@@ -362,3 +363,188 @@ newUserApplicationPre: {
...
+telnetConnected: {
+    art: CONNECT
+    next: clientSelectEncoding
+    config: { nextTimeout: 1500 }
+}
+
+clientSelectEncoding: {
+    art: CLTSEL.ASC
+    next: matrix
+    config: { interrupt: realtime }
```
NuSkooler, line 388: The interrupt probably doesn't buy much here since this would be a unauthenticated user -- with that said, we probably *don't* want any sorts of interruptions here if that were possible.

docs/_docs/configuration/menu-hjson.md
```diff
@@ -362,3 +363,188 @@ newUserApplicationPre: {
     // note that the rest of this menu is omitted for clarity
 }
 ```
+
+## Case Study: Manual Encoding Selection
+
+Enigma½ tries to automatically determine the proper encoding for a client when it connects. Unfortunately, there are cases where the wrong encoding can be selected, resulting in terminal programs that are not supported. If your user base contains users that would like to connect with unsupported clients, one solution is to offer manual encoding selection.
+
+A new system method has been added to support this, @systemMethod:setClientEncoding.
```
NuSkooler, line 371: nit: Maybe not say new here as that would go stale. Perhaps a mention in WHATSNEW though.

docs/_docs/configuration/menu-hjson.md
```diff
@@ -362,3 +363,188 @@ newUserApplicationPre: {
     // note that the rest of this menu is omitted for clarity
 }
 ```
+
```
NuSkooler, line 366: Love all your doc work!!!

cognitivegears:
> With just the one system method alone, one could make a `choose_encoding.js` mod for example that would toggle between UTF-8 and CP438 (/ others) visually like some of the boards do. Lots of possibilities.

Exactly I am hoping to do that, though maybe slightly more generic because in the future I'd like to do the same thing for choosing between formatting (including a checkbox for ANSI/No ANSI, but could also include similar checkboxes for RIP, SIXEL, etc) as well, so I was thinking about implementing it as `terminal_option.js` or something like that which could handle either or both. As part of it I was also planning to make an `art_view.js` as well. That way it could be used for the visual toggle like you are talking about, but in addition could also be used on other screens (it would look cool to have a dynamic main menu, BBS door games that show a preview as you scroll through a list, etc... lots of potential uses.) In order to do this, there is an existing option to `forceLineTerm`. I'm thinking we can extend that with a new option so that instead of adding newlines, it instead adds `ansi.goto()` statements to the next row/starting column. This would allow the art to be even in a "frame" along with other art.

docs/_docs/configuration/menu-hjson.md
cognitivegears: Hmm, so I need to play with that a bit more. I tried that, and it seems like it wouldn't handle at all in that case, though honestly I probably just had a silly mistake in the syntax or something, I'll revisit that thanks.

docs/_docs/configuration/menu-hjson.md
cognitivegears: Thanks, I was planning on changing this again in the future to add other ones (I'd really like to add conversions for at least PETSCII), but good point that others are already technically supported, I'll change the wording for now.

NuSkooler:
FYI you certainly can already make dynamic menus and even base them off encoding or any other ACS.

Are you thinking something like the `show_art.js` module?

BTW, I have a library that I started (but haven't had time to hack on in quite a while) that wraps up various BBS/terminal related tech, detection, so on that I plan on someday using here. You can peek at some of it here: https://github.com/NuSkooler/ansi-bbs-utils

Take a look at capabilities for example: https://github.com/NuSkooler/ansi-bbs-utils/blob/master/lib/capabilities.js

docs/_docs/configuration/menu-hjson.md
NuSkooler: iconv can be extended -- and in fact, I still need to patch it to properly convert UTF-8 to CP437 for *BBS purposes* where a couple control characters are currently stripped (where BBS's will display them generally)

Anyway, should be able to put in PETSCII mapping into iconv and make that avail.

core/system_menu_method.js
cognitivegears: I'd be happy to do that, but I'm not sure what to do here besides log?

core/system_menu_method.js
cognitivegears: Fixed, thanks.

docs/_docs/configuration/menu-hjson.md
cognitivegears: Changed. And definitely later additional encodings is something I'd love to tackle.

docs/_docs/configuration/menu-hjson.md
cognitivegears: Done. Ok I totally didn't know you could do that :) I went back and looked, and there is a comment in the documentation about this but I also added a bit more to hopefully make it obvious to other people as well, because that's a good feature.

docs/_docs/configuration/menu-hjson.md
cognitivegears: Removed the interrupt, thanks!

docs/_docs/configuration/menu-hjson.md
cognitivegears: Done, thanks.

docs/_docs/configuration/menu-hjson.md
cognitivegears: Thanks! As I tell people at work, it doesn't matter if you do something if you don't tell people about it :D I always think updating docs is a big priority, and appreciate all the help you've given on it!

cognitivegears:
> FYI you certainly can already make dynamic menus and even base them off encoding or any other ACS.
> 
> Are you thinking something like the `show_art.js` module?
> 
> BTW, I have a library that I started (but haven't had time to hack on in quite a while) that wraps up various BBS/terminal related tech, detection, so on that I plan on someday using here. You can peek at some of it here: https://github.com/NuSkooler/ansi-bbs-utils
> 
> Take a look at capabilities for example: https://github.com/NuSkooler/ansi-bbs-utils/blob/master/lib/capabilities.js

That's super interesting, I'll dig into that a bit more. As far as the artwork, I think it's a little different than the show_art.js module. I'm thinking about a view component that could go into any menu, that could show multiple artwork inside another page (something like dynamic content inside a div or frame if this were html.) And I thought the trigger for changing could be an event, in particular the `index update` event from `menu_view.js`. That way as items are selected inside a menu it could switch art. Though it would be even better if this was configurable to other events, etc.

core/system_menu_method.js
NuSkooler: I think you're right, not much to do here. Since the client_term already checks, we can just keep it there and keep checking there as well.

NuSkooler:
This LGTM, any last updates before it goes in?

cognitivegears:
> This LGTM, any last updates before it goes in?

Nope, thanks for the help!

PR #461, opened 2022-12-31
Added additional fields from Masto

cognitivegears:
This just adds some additional fields from Mastodon that I'm pretty sure are needed to get even an initial follow going. Not sure on the URLs btw - if you prefer to use something different these can be changed.

### DIFF
--- a/core/servers/content/web_handlers/webfinger.js
+++ b/core/servers/content/web_handlers/webfinger.js
@@ -6,7 +6,6 @@ const WebServerPackageName = require('../web').moduleInfo.packageName;
 
 const _ = require('lodash');
 const User = require('../../../user');
-const { result } = require('lodash');
 const Log = require('../../../logger').log;
 
 exports.moduleInfo = {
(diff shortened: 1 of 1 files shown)


core/servers/content/web_handlers/webfinger.js
```diff
@@ -77,17 +77,41 @@ exports.getModule = class WebFingerServerModule extends ServerModule {
...
+        const href = this._profileUrl(user);
         return {
             rel: 'http://webfinger.net/rel/profile-page',
             type: 'text/plain',
             href,
         };
     }
 
+    _selfUrl(user) {
+        return this.webServer.instance.buildUrl(`/users/${user.username}`);
```
NuSkooler, line 94: We'll need to decide where we want all these URLs to point to, and how much should be configurable. Since the web server can be used for arbitrary hosting, we'll want some of these to be less generic I think. 

I recently added `/_internal/` and `/_f/`, and `/wf/` above. I wonder if `_internal` should be renamed to e.g. `/_enig/`, and we stuff all the rest under that:
* `/_enig/pw/`: password resets
* `/_enig/2fa/`: 2FA/OTP stuff
* `/_enig/wf/`: WebFinger stuff
* `/_enig/ap/`: ActivityPub related stuff

tracker1: Maybe `/enigma-api/*` - Assuming ActivityPub isn't a standardized endpoint.

PR #476, opened 2023-05-21
Rumorz display bugfix

parkbanks:
Noticed the rumorz module was always displaying the first "N" entries, rather than the "N" latest entries. Updated ordering to display the N latest rumors, so that the newest entries are always visible.


NuSkooler:
@parkbanks Awesome, thanks for the fix!

NuSkooler (approved):

PR #477, opened 2023-06-19
Bugfix to theme.hjson

parkbanks:
Changed `messageBaseSearchMessageList` to `messageBaseSearchResultsMessageList` to match `message_base.in.hjson` (line 365) as %VM1 in `MSRCHLST.ANS` was not rendering properly.


NuSkooler (approved):
Looks good, thanks!!

PR #480, opened 2023-07-29
Make real name and other properties optional - pass 1

NuSkooler:
This is intended to allow various fields in the default `nua.js` and their associated user properties as optional.

User properties such as 'real name' are simply stored as `''` if not set. Ops can decide if they want to require these fields in the application process or not.


core/web_password_reset.js
```diff
@@ -143,7 +143,7 @@ class WebPasswordReset {
                     }
 
                     const message = {
-                        to: `${user.properties[UserProps.RealName] || user.username} <${
+                        to: `${user.realName()} <${
```
cognitivegears, line 146: See above - also, maybe a utility method to format an email address would be handy?

cognitivegears (approved):
Looks good to me!

Issue #442, opened 2022-07-29
Some ANSI files do not display correctly on wide screens


cognitivegears:
**Describe the Bug**
When using ANSI files created on some editors it appears that they do not display correctly when the terminal width is wider than the image. I've sen this with multiple files, though not all - it looks likely that it is files that are saved exactly at 80 columns width.

**To Reproduce**
Unknown exactly. I believe it may be that some ANSI editors (I mostly used Moebius, though I feel like this could happen on others as well depending on settings.) My theory is that this happens when an editor assumes the terminal width will be the same as the ANSI width and does not include CR/LF, though that is just a guess on my part currently.

**Expected Behavior**
All common ANSI art should display correctly regardless of wider terminal widths.

**Actual Behavior**
See attached screenshot as well as ANSI file itself.

**Screenshots**

The way it looks with Alacritty:

![MATRIX](https://user-images.githubusercontent.com/3256153/181826759-06f6c13b-db5c-427c-a66c-75aa5e2a0979.png)

Original ANS file:

Unfortunately I am unable to attach an `.ANS` file, please see below for the base64 version. You can save this to a file and use `base64 -d` to decode:

```
G1swbSAbWzFt2sTExMTExMTExMTExMTExMTExMTExMTExMTExMTExMTExMTExMTExMTExMQbWzA7
MzZtxMTExMTExMTExBtbMTszMG3ExMTExMTExMQbWzM0bcTExMTExMTExMTExMTEvxtbMG0NChtb
MTszNG0gG1szN22zG1swbSAgICAgICAgICAgICAgICAgICAgICAgICAgICAgICAgICAgICAgICAg
ICAgICAgICAgICAgICAgICAgICAgICAgICAgICAgICAgICAbWzE7MzRtsxtbMzBtvyAbWzM3bbMg
INrcxMTEG1szNm3ExBtbMG3ExBtbMTszNG3EvxtbMzNt2tzExMQbWzMybcTEG1swbcQbWzMybcTE
vxtbMTszN23a3MTEG1szNm3ExBtbMG3ExBtbMzZtxL8bWzM3bSAgICAgIBtbMW3a3BtbMzVtxL8b
WzM3bdrcxBtbMzZtxMQbWzBtxBtbMzZtxMTEvxtbMTszM23a3MTEG1szMm3EG1swbcQbWzMybcTE
[...]

PR #488, opened 2023-08-25
Fix to display ANSI files with implicit new lines on wide terminals

cognitivegears:
Some art programs can save ANSI without carriage returns to save a few bytes in the file. This works fine when the terminal width matches the ANSI (since most retro is 80 columns) but on newer terminals, etc, this massively breaks the display.

The issue is discussed in issue #442 .  In order to fix this, we force a newline if the SAUCE is present, newlines are missing in the art file, and the width is less than the terminal width.

### DIFF
--- a/core/ansi_escape_parser.js
+++ b/core/ansi_escape_parser.js
@@ -37,6 +37,10 @@ function ANSIEscapeParser(options) {
     this.mciReplaceChar = miscUtil.valueWithDefault(options.mciReplaceChar, '');
     this.termHeight = miscUtil.valueWithDefault(options.termHeight, 25);
     this.termWidth = miscUtil.valueWithDefault(options.termWidth, 80);
+    this.breakWidth = this.termWidth;
+    if(!(_.isNil(options.artWidth)) && options.artWidth > 0 && options.artWidth < this.breakWidth) {
+        this.breakWidth = options.artWidth;
+    }
     this.trailingLF = miscUtil.valueWithDefault(options.trailingLF, 'default');
 
     this.row = Math.min(options?.startRow ?? 1, this.termHeight);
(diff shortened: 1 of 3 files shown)


core/ansi_escape_parser.js
```diff
@@ -37,6 +37,10 @@ function ANSIEscapeParser(options) {
     this.mciReplaceChar = miscUtil.valueWithDefault(options.mciReplaceChar, '');
     this.termHeight = miscUtil.valueWithDefault(options.termHeight, 25);
     this.termWidth = miscUtil.valueWithDefault(options.termWidth, 80);
+    this.breakWidth = this.termWidth;
+    if(!(_.isNil(options.artWidth)) && options.artWidth > 0 && options.artWidth < this.breakWidth) {
```
NuSkooler, line 41: We should probably use `parseInt()` and `isNaN()` here on `artWidth`.

core/ansi_escape_parser.js
```diff
@@ -105,22 +109,25 @@ function ANSIEscapeParser(options) {
...
                     self.positionUpdated();
                     break;
 
                 default:
-                    if (self.column === self.termWidth) {
+                    if (self.column === self.breakWidth) {
                         self.emit('literal', text.slice(start, pos + 1));
                         start = pos + 1;
 
+                        // If the art is terminal, then we need to force the terminal to go to the next line.
```
NuSkooler, line 125: I'm having trouble understanding the comment (though I understand the code)

NuSkooler:
This is going to be awesome to have fixed!

core/ansi_escape_parser.js
cognitivegears: I a word. Fixed :)

core/ansi_escape_parser.js
cognitivegears: I had done that in art.js but your right, shouldn't leave it to that. Updated.

NuSkooler (approved):

PR #498, opened 2023-09-05
Bugfix/fix art length

cognitivegears:
Changed a few art files that had an art length of 25 rows to 24 (as well as any associated theme settings to make them work.) Also fixed a few files with bad SAUCE info. Then did a *ton* of testing of terminals, including SyncTerm, MagiTerm, and Icy Term for the CP437 terminals, along with UTF-8 terminals Windows Terminal, Alacritty, Hyper.js, ConEMU, and PuTTY. Funnily enough, PuTTY actually looked the best of the utf-8 terms, though all of them displayed cleanly (except for some color issues that were not the fault of Enigma.) All were usable however. SyncTerm with it's border at row 25 also worked great on all screens without scrolling.


NuSkooler:
Thank you for this!

NuSkooler (approved):

PR #501, opened 2023-09-19
fixed incorrect reporting of cursor position

AnthonyHarwood:
Changed the string padding so the leading zero will be affixed to the start of the string, rather than the end of the string.


cognitivegears (approved):
Tested and verified. Thanks for the bugfix!

NuSkooler (approved):

PR #503, opened 2023-09-22
Added code to clear the text

AnthonyHarwood:
I noticed the text in the error view was not being cleared after the error was corrected.  For example, "Username is too short". 

I can think of a few different ways this could have been fixed.  I try to keep it simple.  If you prefer a different approach to resolving this, please don't hesitate to let me know.

Thanks!

### DIFF
--- a/core/text_view.js
+++ b/core/text_view.js
@@ -179,6 +179,10 @@ TextView.prototype.setText = function (text, redraw) {
 };
 
 TextView.prototype.clearText = function () {
+    if (this.text) {
+        this.setText(this.fillChar.repeat(this.text.length));
+    }
+
     this.setText('');
 };
 


core/text_view.js
cognitivegears, line 1: This looks like a great change! See below for one small comment -

core/text_view.js
```diff
@@ -179,6 +179,10 @@ TextView.prototype.setText = function (text, redraw) {
 };
 
 TextView.prototype.clearText = function () {
+    if (this.text) {
+        this.setText(' '.repeat(this.text.length));
```
cognitivegears, line 183: Can we use `this.fillChar` here instead of a literal space?

AnthonyHarwood (commented):
Good suggestion.  Thanks.

cognitivegears (approved):
Looks good to me, I'll wait for @NuSkooler in case he has any concerns but thanks so much for the PR, keep 'em coming :)

NuSkooler (approved):
Awesome, I like simple! Thanks!

PR #505, opened 2023-09-26
Fix for mail to invalid FTN address

AnthonyHarwood:
Added a handler to catch invalid FTN address, that was closing socket connection.

### DIFF
--- a/core/system_view_validate.js
+++ b/core/system_view_validate.js
@@ -7,6 +7,7 @@ const Config = require('./config.js').get;
 const Log = require('./logger.js').log;
 const { getAddressedToInfo } = require('./mail_util.js');
 const Message = require('./message.js');
+const Errors = require('./enig_error.js').Errors;
 
 //  deps
 const fs = require('graceful-fs');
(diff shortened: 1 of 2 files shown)


AnthonyHarwood:
To reproduce the issue, create a mail message and address it to 555

cognitivegears:
@NuSkooler can you test this out? I don't have any fidonet networks configured.

core/system_view_validate.js
```diff
@@ -83,19 +83,23 @@ function validateUserNameOrRealNameExists(data, cb) {
...
-    //  :TODO: remove hard-coded FTN check here. We need a decent way to register global supported flavors with modules.
-    const addressedToInfo = getAddressedToInfo(data);
+    try {
+        //
+        //  Allow any supported addressing:
+        //  - Local username or real name
+        //  - Supported remote flavors such as FTN, email, ...
+        //
+        //  :TODO: remove hard-coded FTN check here. We need a decent way to register global supported flavors with modules.
+        const addressedToInfo = getAddressedToInfo(data);
```
NuSkooler, line 93: Looks like we're just catching a `undefined/null`. This can probably be switched to a less heavy `if (!addressedToInfo) { }`.

For the error itself, please use [Errors.Invalid](https://github.com/NuSkooler/enigma-bbs/blob/master/core/enig_error.js#L37), e.g. `Errors.Invalid('Invalid address')`

NuSkooler:
Thanks for catching this! I left a small comment for a tidy, but should be good to go after

core/system_view_validate.js
AnthonyHarwood: if (!addressedToInfo) is much better. Good catch.

Regarding your other suggestion... I actually _had_ Errors.Invalid('address') in my initial commit, until I realized it produced "Invalid: address" while the other error messages in the header-- "Invalid username", "Subject too short", were not using Errors.Invalid(), so they didn't look the same.  So I got rid of Errors.Invalid() to try and be consistent with the other existing error messages, especially "Invalid username", since it does not have the colon in the error message.

I'll be glad to make the change to use Errors.Invalid(), and if you prefer I can also make the same change to "Invalid username" so it will become "Invalid: username", that is if you want the format of error messages to match.  But "Subject too short" will need to remain as-is since "Invalid: Subject too short" is too lengthy to fit in the space allocated in the blocktronics theme.

Thanks for your feedback, and your patience with me as I try to improve my JSing. I'll wait for your next feedback before I make the changes you have suggested. And I apologize if I am overthinking this.  I'm glad to be contributing, and I want to do a good job.  Thanks for the opportunity :)

AnthonyHarwood:
Closing this and submitting a simpler solution.

PR #507, opened 2023-09-26
Removed extraneous space from private mail config template

AnthonyHarwood:
An extra space in the template was preventing the escape key from functioning within the private mail message header.  This will fix the issue for new installations.  Existing sysops would need to modify their config, since it has already been copied from the template. No biggie.


cognitivegears (approved):
Wow, thanks for the great catch!!

PR #510, opened 2023-09-26
Bugfix/scrolling updates

cognitivegears:
This fixes issues with scrolling updates and MCI codes, as well as a bunch of additional ANSI positioning and compatibility fixes to increase interoperability with various ANSI drawing programs and terminals.


NuSkooler (approved):
Looks great! Makes me even more want to come in and wipe out those regex's and just make it all a simpler state machine.

PR #515, opened 2023-10-11
Updated the SSH version, config, and documentation

cognitivegears:
This ended up being a bit bigger than I had initially intended, as I was just going to update documentation. However, as I dug into it there was a number of things out of date or that I thought could use a cleanup. I made the following changes:

* Updated the version of ssh2 to 1.14
* Updated the algorithms - there were some algorithms that didn't work with newly generated keys with new versions of ssh-keygen, and some that we could add now because we have a newer version of node.
* Updated lots of documentation, including troubleshooting :)

Note: all of this has been tested on new terminals, netrunner, icyterm, magiterm, and syncterm and look like it's running great.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,5 @@ mail/
 node_modules/
 docs/_site/
 docs/.sass-cache/
+
+docs/.jekyll-cache/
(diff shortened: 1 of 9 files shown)


melroy89:
Might fix https://github.com/NuSkooler/enigma-bbs/issues/518 ?

NuSkooler (approved):
Looks great, and seems to be working great as well. NetRunner issue I had with a previous update seems to have been fixed.

Awesome work!

PR #520, opened 2023-10-15
Feature/multiarch docker

cognitivegears:
I made a few improvements to the Docker build process, including:

* Upgrading to node v20
* Upgraded the qemu and buildx action versions
* Reordered some of the Dockerfile to avoid snapshot changes on each code change
* Updated the image to use bookworm instead of buster

These changes appear to also have fixed the issue with networking on the build along the way.  See build (https://github.com/NuSkooler/enigma-bbs/actions/runs/6526380735) for a successful run.

*Note* I have not tried running these docker containers on real hardware yet, this needs to be tested.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -12,7 +12,7 @@ ENV DEBIAN_FRONTEND noninteractive
 
 
 # Just copy the package.json so it only needs to build once
-COPY package.json /enigma-bbs
+COPY package.json /enigma-bbs/
 
 # Install APT and NPM packages
 RUN apt-get update \
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -20,7 +20,6 @@ RUN apt-get update \
     git \
     curl \
     build-essential \
-    python \
     python3 \
     libssl-dev \
     lrzsz \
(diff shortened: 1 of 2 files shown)


NuSkooler (approved):

docker/Dockerfile
```diff
@@ -22,26 +28,33 @@ RUN apt-get update \
...
+    && cd /enigma-bbs \
     && pm2 start main.js \
     && mkdir -p /enigma-bbs-pre/art \
     && mkdir /enigma-bbs-pre/mods \
     && mkdir /enigma-bbs-pre/config \
     && cp -rp art/* ../enigma-bbs-pre/art/ \
     && cp -rp mods/* ../enigma-bbs-pre/mods/ \
     && cp -rp config/* ../enigma-bbs-pre/config/ \
-    && apt-get remove build-essential python python3 libssl-dev git curl -y \
+    && apt-get remove build-essential python3 libssl-dev git curl -y \
```
tracker1, line 53: Maybe consider a multi-stage build/runtime, as this doesn't decrease the actual download size for image slices.

Issue #522, opened 2023-11-17
ENIGMA crashes if NUA does not collect complete info.


chilledtonic:
**Describe the Bug**
My understanding was that [this pr](https://github.com/NuSkooler/enigma-bbs/pull/480) made collecting full user information unnecessary; however, if you later open the user config page after having some of these values unset (on my board, I do not collect real name or location), and try to tab past an empty line in the config, the board will hard crash with this error.

```
enigma-bbs/core/text_view.js:156
        text = text.toString();  
```

**To Reproduce**
1. Establish a board where NUA does not collect name or location (or skip any default field)
2. Try to edit the account

**Expected Behavior**

I would prefer either ENIGMA more gracefully handled the empty string.

**Actual Behavior**

Server crashes completely.

**Screenshots**

**Environment**
* [X] I am using Node.js v14.x LTS or higher
* [x] `npm install` or `yarn` reports success

NuSkooler:
@chilledtonic thanks for the report & PR! I should be able to take a look this week.

voipmeister:
Can reproduce on a fresh installation of 0.0.14-beta on a RPi. Going through the application form crashes Enigma on using Tab or Enter. This is the feedback on the console:

`/mybbs/enigma-bbs/core/text_view.js:156
        text = text.toString();
                    ^

TypeError: Cannot read property 'toString' of undefined
    at TextView.setText (/mybbs/enigma-bbs/core/text_view.js:156:21)
    at viewValidationListener (/mybbs/enigma-bbs/core/nua.js:56:32)
    at validateResult (/mybbs/enigma-bbs/core/view_controller.js:395:21)
    at EditTextView.validateNonEmpty [as validate] (/mybbs/enigma-bbs/core/system_view_validate.js:27:12)
    at ViewController.validateView (/mybbs/enigma-bbs/core/view_controller.js:387:18)
    at ViewController.switchFocus (/mybbs/enigma-bbs/core/view_controller.js:525:10)
    at ViewController.nextFocus (/mybbs/enigma-bbs/core/view_controller.js:553:14)
    at EditTextView.ViewController.viewActionListener (/mybbs/enigma-bbs/core/view_controller.js:100:22)
    at EditTextView.emit (events.js:400:28)
    at EditTextView.View.onKeyPress (/mybbs/enigma-bbs/core/view.js:310:18)`

NuSkooler:
[...]

PR #523, opened 2023-11-17
🐛 bug: Server no longer crashes if CONFSCR has null values

chilledtonic:
This Bugfix corrects issue #522 by doing a check for null values anywhere that gets passed the raw values of the (potentially null) config screen.

This works for the bug in question, but I don't have enough experience in this code base to identify if this may have unintended consequences. Let me know if there is anything I've missed and I'll revise my commits.

### DIFF
--- a/core/color_codes.js
+++ b/core/color_codes.js
@@ -98,7 +98,7 @@ function ansiSgrFromCnetStyleColorCode(cc) {
 }
 
 function renegadeToAnsi(s, client) {
-    if (-1 == s.indexOf('|')) {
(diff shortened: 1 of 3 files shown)


cognitivegears:
Thanks so much @chilledtonic ! This is fantastic. Just a small possible change, I'm wondering if we should use lodash since we have it here, just to capture undefined, etc as well (either isNil or isEmpty depending on context.)

Really appreciate the submission!

NuSkooler:
Thanks for your submission! I made a slightly different change -- it should resolve the issue

PR #533, opened 2024-03-19
Fix for x-forwarded-for not being applied

notepid:
this.proxied is not set before calling setClient() resulting in a "undefined" value for this.proxied. This in turn prevents the proxy headers to be used when resolving the client IP.

In a scenario where a reverse proxy is used the client IP will always show as 127.0.0.1 instead of the X-Forwarded-For header.

Also consider using request.socket instead of connection in setClient() Ref: https://stackoverflow.com/questions/8107856/how-to-determine-a-users-ip-address-in-node

### DIFF
--- a/core/servers/login/websocket.js
+++ b/core/servers/login/websocket.js
@@ -71,6 +71,25 @@ class WebSocketClient extends TelnetClient {
         })(ws);
 
         super(wsDuplex);
+
+        Log.trace({ headers: req.headers }, 'WebSocket connection headers');
+
+        //
+        //  If the config allows it, look for 'x-forwarded-proto' as "https"
+        //  to override |isSecure|
+        //
+        if (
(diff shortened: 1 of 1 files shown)


NuSkooler:
@notepid Thanks for this! 

RE: request.socket, is the current method causing issues?

notepid:
> @notepid Thanks for this!
You're welcome :) 

> RE: request.socket, is the current method causing issues?

No issues as I see. Just noted while looking into this that connection is being deprecated.

I tested with this code and it looked to work but that was just a quick test:
`this.resolvedRemoteAddress =
(this.client.proxied && httpRequest.headers['x-forwarded-for'] || '').split(',').pop().trim() || httpRequest.socket.remoteAddress;`

Apparently x-forwarded-for can be a list of addresses.

NuSkooler (approved):

PR #545, opened 2024-07-20
Added SSL support

stack-fault:
Added SSL support between multiplexer and server
Updated templates with new configuration options

### DIFF
--- a/misc/config_template.in.hjson
+++ b/misc/config_template.in.hjson
@@ -322,7 +322,7 @@
             serverHostname  : 'mrc.bottomlessabyss.net'
             serverPort      : 5000
             serverSslPort   : 5001
-            useSsl          : false
+            useSsl          : true
             retryDelay      : 10000
             multiplexerPort : 5000
         }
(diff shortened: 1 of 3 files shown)


NuSkooler (approved):
Looks great, thanks @stack-fault !

misc/config_template.in.hjson
```diff
@@ -318,9 +318,13 @@
...
 
         mrc: {
-            enabled             : false
-            serverHostname      : 'mrc.bottomlessabyss.net'
-            serverPort          : 5000
+            enabled         : false
+            serverHostname  : 'mrc.bottomlessabyss.net'
+            serverPort      : 5000
+            serverSslPort   : 5001
+            useSsl          : false
```
NuSkooler, line 325: Is there any reason not to use TLS by default?

stack-fault: I usually leave it disabled because in some case it requires an optional library not always installed.
If that is not the case here, please set it as default to true for sure.

NuSkooler: Awesome, updated. Should be fine since Node has TLS and that's all that's really happening in this case

NuSkooler:
Thanks again @stack-fault !

Issue #551, opened 2024-08-24
mrc_multiplexer.js errors out when URL is pasted in MRC chat


voipmeister:
**Describe the Bug**
Enigma crashes when a URL is pasted in the MRC chat.

**To Reproduce**
Log onto MRC from Enigma 0.0.14, have someone paste for instance: `https://twitch.tv/the_meatlotion`, from a Mystic board (probably any UTF-8 client). In this case it was Meatlotion.

**Expected Behavior**
Proper sanitation, no crash. Other chatters had no issue in the same chat, but they use different clients obviously.

**Actual Behavior**
Enigma crashes:
```
ENiGMA½ Copyright (c) 2015-2024, Bryan D. Ashby
_____________________   _____  ____________________    __________\_   /
/____       _____|      __________       ___|__|      ____|     \   /  _____  \
---- \______\ -- |______\ ------ /______/ ---- |______\ - |______\ /__/ // ___/
                                                                     /__   _\
 <*>   ENiGMA½  // HTTPS://GITHUB.COM/NUSKOOLER/ENIGMA-BBS   <*>       /__/

-------------------------------------------------------------------------------

System started!
/home/enigma/enigma-bbs/core/servers/chat/mrc_multiplexer.js:249
                message.to_user.toUpperCase() == client.username.toUpperCase() ||
                                ^

TypeError: Cannot read properties of undefined (reading 'toUpperCase')
    at /home/enigma/enigma-bbs/core/servers/chat/mrc_multiplexer.js:249:33
    at Set.forEach (<anonymous>)
    at MrcModule.sendToClient (/home/enigma/enigma-bbs/core/servers/chat/mrc_multiplexer.js:245:26)
    at MrcModule.receiveFromMRC (/home/enigma/enigma-bbs/core/servers/chat/mrc_multiplexer.js:323:18)
    at /home/enigma/enigma-bbs/core/servers/chat/mrc_multiplexer.js:157:30
    at Array.forEach (<anonymous>)
    at handleData (/home/enigma/enigma-bbs/core/servers/chat/mrc_multiplexer.js:153:19)
    at TLSSocket.<anonymous> (/home/enigma/enigma-bbs/core/servers/chat/mrc_multiplexer.js:164:13)
    at TLSSocket.emit (node:events:517:28)
    at addChunk (node:internal/streams/readable:368:12)

Node.js v18.20.4
[enigma@node01 enigma-bbs]$
```

**Screenshots**
N/a.

**Environment**
[...]

PR #552, opened 2024-08-27
Fixed issue on URL resolver message

stack-fault:
Moved user_to uppercase to a different function
This fixes issue #551

### DIFF
--- a/core/mrc.js
+++ b/core/mrc.js
@@ -47,7 +47,7 @@ const MciViewIds = {
 const helpText = `
 |15General Chat|08:
 |03/|11rooms |08& |03/|11join |03<room>      |08- |07List all or join a room
-|03/|11pm |03<user> <message>       |08- |07Send a private message
+|03/|11pm |03<user> <message>       |08- |07Send a private message |08(/t /tell /msg)
 ----
 |03/|11whoon                     |08- |07Who's on what BBS
 |03/|11chatters                  |08- |07Who's in what room
@@ -57,6 +57,7 @@ const helpText = `
 |03/|11meetups                   |08- |07Info about MRC MeetUps
 |03/|11quote                     |08- |07Send raw command to server
 |03/|11help                      |08- |07Server-side commands help
+|03/|11quit                      |08- |07Quit MRC |08(/q)
 ---
 |03/|11l33t |03<your message>       |08- |07l337 5p34k
 |03/|11kewl |03<your message>       |08- |07BBS KeWL SPeaK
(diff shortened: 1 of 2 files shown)


core/servers/chat/mrc_multiplexer.js
```diff
@@ -245,8 +245,7 @@ exports.getModule = class MrcModule extends ServerModule {
         connectedSockets.forEach(client => {
             if (
                 message.to_user == '' ||
-                // Fix PrivMSG delivery on case mismatch
-                message.to_user.toUpperCase() == client.username.toUpperCase() ||
+                message.to_user == client.username.toUpperCase() ||
```
NuSkooler, line 248: These would technically be better using [localeCompare](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare) with sensitivity 'accent'.

core/servers/chat/mrc_multiplexer.js
```diff
@@ -336,6 +335,13 @@ exports.getModule = class MrcModule extends ServerModule {
         //     return;
         // }
 
+        // Make sure to_user is always uppercase
+        try {
+            let to_user_uc = to_user.toUpperCase();
+            to_user = to_user_uc;
```
NuSkooler, line 341: This could just be `to_user = to_user.toUpperCase()`, no temp var. Or without the try:
```javascript
to_user = (to_user || '').toUpperCase()
```

stack-fault:
Ok, I did a couple things in here to fix things I noticed while testing.
I will open another issue for something that is buried down elsewhere for which you may have an idea.

stack-fault:
@NuSkooler let me know if you have additional questions before merging the PR

core/mrc.js
```diff
@@ -275,21 +275,7 @@ exports.getModule = class mrcModule extends MenuModule {
...
-                padAmount = chatWidth - (messageLength % chatWidth) - spaces;
-            } else {
-                padAmount = chatWidth - messageLength - spaces;
-            }
-
-            if (padAmount < 0) padAmount = 0;
-
-            const padding = ' |00' + ' '.repeat(padAmount);
-            chatLogView.addText(pipeToAnsi(msg + padding));
+            chatLogView.addText(pipeToAnsi(msg));
```
NuSkooler, line 278: I'm not sure this will work correctly, I'll have to do some testing. This MR just got a bit biggerr

stack-fault: I have the latest client and it worked fine on my end. All the padding appeared to be from an older way of drawing the screen which appeared to have changed since it was first implemented (assumption here). On my end, it was working pretty well without it and fixed an issue with text wrapping on certain cases. Let me know if there is anything I missed.

stack-fault:
@NuSkooler
Pushed some additional small fixes to the branch since it's not merged yet.

core/mrc.js
```diff
@@ -657,21 +692,26 @@ exports.getModule = class mrcModule extends MenuModule {
...
      * Joins a room, unsurprisingly
      */
-    joinRoom(room) {
+    async joinRoom(room) {
         // room names are displayed with a # but referred to without. confusing.
         room = room.replace(/^#/, '');
         this.state.room = room;
         this.sendServerMessage(`NEWROOM:${this.state.room}:${room}`);
+
+        await this.msgDelay(100);
```
NuSkooler, line 701: nice!

NuSkooler (approved):
Looks good to me, thanks for your work on this!

PR #553, opened 2024-08-27
Update README.md - Magiterm

chilledtonic:
Updated README reference to Magiterm, removing old dead link and instead listing the most recent development build on their Gitlab.


NuSkooler (approved):
Thank you for this!

PR #564, opened 2024-09-01
Continue parsing pkt if bad message

NuSkooler:
Continue (attempting) to parse messages from a PKT if one is deemed bad; log errors


NuSkooler:
Not really needed

PR #565, opened 2024-09-01
Fix: /bbses, /identify, /register, /update

christiansacks:
Fixed;

/bbses
The server accepts args to filter out the list, now you can do that in Enigma with this fix;
/bbses arg1 arg2 arg3 argn

Added for ease of ident use;
/identify
/register
/update

### DIFF
--- a/core/mrc.js
+++ b/core/mrc.js
@@ -527,7 +527,8 @@ exports.getModule = class mrcModule extends MenuModule {
                 break;
 
             case 'bbses':
-                this.sendServerMessage('CONNECTED');
+		const args = cmd.slice(1).join(' ');
+                this.sendServerMessage(`CONNECTED ${args}`);
                 break;
 
             case 'topic':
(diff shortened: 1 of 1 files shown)


stack-fault:
/identify, /register and /update have been added in a previous PR, this is now a duplicate in this one.

voipmeister:
I tested the raw `mrc.js` on my installation with all of the commands above, except for `/register`.

All of them work 👍 

Filtering with `/bbses don` returns too many rows, but this is the case with the Mystic client as well.

NuSkooler:
Hrm, there are 94 files changed here. Probably picked up more than you wanted?

christiansacks:
> Hrm, there are 94 files changed here. Probably picked up more than you wanted?

That was an oopsie!! I have removed the erroneous commit. Feel free to merge if you will be so kind sir.

stack-fault:
As a sidenote, most of your PR Christian is already in the code. Please check below for a section called MRC Trust I had pushed last week.

PR #566, opened 2024-09-08
Addressing #563 (Cannot parse privateKey: Unsupported key format)

voipmeister:
Changed config template + docs to address https://github.com/NuSkooler/enigma-bbs/issues/563


NuSkooler (approved):
Awesome, thank you!

PR #573, opened 2024-11-21
Update install.sh to default to Node v18

crhultay:
Manual Installation per https://nuskooler.github.io/enigma-bbs/installation/manual.html asks the user to use v18 whereas this script was defaulting to 14.

This caused me some frustration in getting my BBS up and running as I at first depended on the auto-install script.

For the less apt user, this might be a blocker to adopting this BBS Software.

### DIFF
--- a/misc/install.sh
+++ b/misc/install.sh
@@ -2,7 +2,7 @@
 
 { # this ensures the entire script is downloaded before execution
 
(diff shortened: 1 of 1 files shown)


NuSkooler (approved):
<3 thank you!

PR #574, opened 2024-11-21
Update bbs.js to for Config Init Enhancements

crhultay:
If a new user of enigma-bbs tries to run ./main.js it will now explain why it can't start and suggest running './oputil.js config new'

In addition, in the event of an error scenario it will exit with exitcode 1 so that systemd / other tools that bring up the service will be able to detect than an error had occurred -- process.exit() defaults to 0, and therefore would not have been caught.

### DIFF
--- a/core/bbs.js
+++ b/core/bbs.js
@@ -85,23 +85,18 @@ function main() {
                     //  then it's a fatal error
                     //
                     if (err) {
+                        errorDisplayed = true;
+                        console.error(`Configuration error: ${err.message}`); //  eslint-disable-line no-console
+
                         if ('ENOENT' === err.code) {
-                            if (configPathSupplied) {
-                                console.error(
-                                    'Configuration file does not exist: ' + configFile
-                                );
-                            } else {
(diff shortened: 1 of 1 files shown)


core/bbs.js
```diff
@@ -85,23 +85,18 @@ function main() {
...
-                        } else {
-                            errorDisplayed = true;
-                            console.error(`Configuration error: ${err.message}`); //  eslint-disable-line no-console
-                            if (err.hint) {
-                                console.error(`Hint: ${err.hint}`);
-                            }
-                            if (err.configPath) {
-                                console.error(`Note: ${err.configPath}`);
-                            }
+                            console.error("\nConfiguration file does not exist: '" + configFile + "'\n\nIf this is a new installation please run './oputil.js config new' from the enigma-bbs directory");
```
NuSkooler, line 92: super nit in case you work on this any more before merged: consistent `{configFile}` style here

NuSkooler (approved):
Thank you again, very much appreciated!

Issue #516, opened 2023-10-11
Bump version of ws library


cognitivegears:
**Describe the Bug**
A moderate vulnerability in the ws library - see https://github.com/advisories/GHSA-6fc8-4gx4-v693

**To Reproduce**
See the above URL for proof of concept code

**Expected Behavior**
Not impacted

Upgrade the ws library to 7.5.6 (or current latest version) and test VTX

crhultay:
Raised https://github.com/NuSkooler/enigma-bbs/pull/577 to fix documentation error that prevented me from testing correctly.

crhultay:
Tested VTX on 8.18.0
![image](https://github.com/user-attachments/assets/26f2cfe2-4955-44c4-97bf-e4e23855906e)

crhultay:
@NuSkooler Could we look at closing this issue? I have confirmed that it indeed works.

I think it's safe to merge PR #557

PR #576, opened 2024-11-23
Issue 516: Bump ws library and test VTX

crhultay:
https://github.com/NuSkooler/enigma-bbs/issues/516

Bumped to 8.18.0 which is the latest

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,7 +65,7 @@
         "temptmp": "^1.1.0",
         "uuid": "8.3.2",
         "uuid-parse": "1.1.0",
-        "ws": "7.4.3",
+        "ws": "8.18.0",
         "yazl": "^2.5.1"
     },
     "devDependencies": {


crhultay:
Tested directly through http
image

Instructions in https://nuskooler.github.io/enigma-bbs/servers/loginservers/websocket.html, and this wants to be simplified into a bash script utility to make it easier for new SysOps.

Setup and tested via VTXClient per instructions in https://nuskooler.github.io/enigma-bbs/servers/loginservers/websocket.html, and this wants to be simplified into a bash script utility to make it easier for new SysOps.

I would see in my logs
```
[2024-11-23T00:48:11.768Z]  INFO: ENiGMA½ BBS/14182 on bbs: Client connected on node 1 (WebSocket/undefined) (nodeId=1, sessionId=W7UdzQD8g5Y, remoteAddress=::ffff:192.168.2.48, friendlyRemoteAddress=192.168.2.48, serverName=WebSocket, isSecure=false)
```

And from HTTP Server provided VTX Client, I see:
image

Telnetted into 192.168.2.12:8810 from another machine, and I get:
image

I'm not sure what is missing, but no errors are being logged by the server and the connection simply times out from the VTXClient whether I'm using ws 7.4.3 or ws 8.18.0.

https://github.com/NuSkooler/enigma-bbs/pull/576 was raised, however I cannot affirm that VTX works or not.

crhultay:
Got it working. The example vtxdata.json was missing an important parameter.

crhultay:
Fixed, and tested on ws 8.18.0
image

NuSkooler (approved):

Issue #519, opened 2023-10-12
Generate SSH cert with oputil


cognitivegears:
Now that we have upgraded the ssh2 library, it supports the ability to generate ssh certificates. We should take advantage of this to facilitate ssh setup without any external tools needed. Some thoughts below:

* Create a new core/oputil/oputil_ssh.js
* allow creation of the key with a random password or manual entered password. Add the password to the config.hjson.
* make sure the private key is saved with the minimum file permissions needed.
* update documentation (current description of running ssh-keygen etc should not be removed but changed to a secondary option.

NuSkooler:
This could probably live under [oputil config](https://github.com/NuSkooler/enigma-bbs/blob/master/core/oputil/oputil_config.js), e.g. `oputil.js config ssh-key [options]`

crhultay:
Raised PR https://github.com/NuSkooler/enigma-bbs/pull/578

Issue #281, opened 2020-05-17
Configuration File Schema Validation


NuSkooler:
It may be nice to have very basic schema validation for configuration files (`config.hjson`, `menu.hjson`, etc.).

* Help users identify issues with their configurations
* Ideally something very light
* Must allow for additional entries/sections that are _not_ known to the schema

Configuration files are in HJSON, but this directly converts to JSON internally. It may be enough to work with the JSON.

JSON schema validation resources:
* https://github.com/yusufnb/verify-json - Looks very interesting in that it is extremely light weight and allows custom validators
* https://json-schema.org/ - This spec is quite intense

See also: #280

louisnorthmore:
This would be useful - would have saved me some time yesterday!

louisnorthmore:
I might take a look at this later. How do you see it working? A separate tool to begin with?

NuSkooler:
@louisnorthmore I haven't put a ton of detailed thought into this, but something like:
* When the configuration is (re)loaded do quick validation
* Possibly an explicit `./oputil.js config validate` if that can provide additional benefits.

I like the idea of veirfy-json since it's light weight and allows custom validators.

NuSkooler:
At this point if it hasn't been picked up yet, I change my stance and suggest JSON schema as it's been the clear winner around JSON... schemas. 

Upon validation error(s), writing to `stderr` and to the log (if possible) should be enough when starting up or hot reloading the config.

cognitivegears:
Agreed with the overall approach, and this would be fantastic. I know when I was getting started having a broken config was my biggest struggle. The only one I'm not sure about is the last bullet point: "Must allow for additional entries/sections that are not known to the schema" - this would somewhat limit some of the usefulness I think? If someone is trying to debug why a configuration item is not taking effect, etc, this would be easier if there was at least the option to restrict to known elements.

NuSkooler:
@cognitivegears I think we can validate things that are in the wrong place, and specifics of all known configuration elements, but consider e.g. a custom mod that is not shipped with enig, but wants some configuration from `config.hjson`. If the user adds a new block **in the right area** of the config, we won't have a way to really validate the inners. Unless of course we expose a way for mods to supply their schemas -- which feels a little gatekeeping as most people won't know what to do there.

In JSON schema at least, this probably mostly around allowing additional properties in certain blocks.

cognitivegears:
[...]

PR #578, opened 2024-11-23
Issue #519 - Implement SSH Key Generation via oputil

crhultay:
Improvement could be to load variables from config
- OpenSSL Binary Path
- SSH Key Path and Filename

I would've done this, but I don't understand how to read configs yet.

### DIFF
--- a/core/oputil/oputil_ssh_key.js
+++ b/core/oputil/oputil_ssh_key.js
@@ -1,0 +1,158 @@
+/* jslint node: true */
+/* eslint-disable no-console */
+'use strict';
+
+//	ENiGMA½
+const initConfigAndDatabases = require('./oputil_common.js').initConfigAndDatabases;
+
+const {
+    printUsageAndSetExitCode,
+    argv,
+    ExitCodes,
+    getAnswers,
+} = require('./oputil_common.js');
+const getHelpFor = require('./oputil_help.js').getHelpFor;
+
+//	deps
+const async = require('async');
+const fs = require('graceful-fs');
+const exec = require('child_process').exec;
+const inq = require('inquirer');
+const _ = require('lodash');
+
+
+exports.handleSSHKeyCommand = handleSSHKeyCommand;
+
+const ConfigIncludeKeys = [
+    'loginServers.ssh',
+    'loginServers.ssh.privateKeyPem',
+];
+
+const MINIMUM_PASSWORD_LENGTH = 8;
+const QUESTIONS = {
+    Create: [
+        {
+            name: 'createNew',
+            message: 'Generate New SSH Keys?',
+            type: 'confirm',
+            default: false,
+        },
+        {
+            name: 'password',
+            message: 'SSL Password:',
+            default: "",
+            when: answers => answers.createNew,
+        },
+    ],
+};
+
+function execute(ui, command) {
+    ui.log.write("Ping!");
+    ui.log.write(command);
+    exec(
+        command,
+        function (error, stdout, stderr) {
+            ui.log.write(error);
+
+            if (error) {
+                const reason = error ? error.message : 'OpenSSL Error';
+                logDebug(
+                    {
+                        reason: reason,
+                        cmd: util.cmd,
+                        args: args
+                    },
+                    `openssl command failed`
+                );
+            }
+            else {
+                ui.log.write("SSH Keys Generated")
+            }
+        }
+    );
+}
+
+function createNew(cb) {
+    const ui = new inq.ui.BottomBar();
+
+    let sslPassword;
+
+    async.waterfall(
+        [
+            function init(callback) {
+                return initConfigAndDatabases(callback);
+            },
+            function create(configuration, callback) {
+                getAnswers(QUESTIONS.Create, answers => {
+                    if (!answers.createNew) {
+                        return callback('exit');
+                    }
+
+                    // Get Answer Value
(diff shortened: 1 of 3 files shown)


core/oputil/oputil_ssh_key.js
```diff
@@ -0,0 +1,158 @@
...
+            },
+            function create(configuration, callback) {
+                getAnswers(QUESTIONS.Create, answers => {
+                    if (!answers.createNew) {
+                        return callback('exit');
+                    }
+
+                    // Get Answer Value
+                    sslPassword = answers.password;
+                    if (!sslPassword || sslPassword.replaceAll(" ", "") == "") {
```
NuSkooler, line 93: You could replace line 92 with `sslPassword = answers.password.trim()` and drop the `replaceAll()` here.

core/oputil/oputil_ssh_key.js
```diff
@@ -0,0 +1,158 @@
...
+                    ui.log.write(`Creating SSH Key: ${targetKeyFile}`);
+
+                    // Create Dir
+                    if (!fs.existsSync(sshKeyPath)) {
+                        ui.log.write(`Creating Directory: ${sshKeyPath}`);
+                        exec(`mkdir -p ${sshKeyPath}`);
+                    }
+
+                    // Check if OpenSSL binary is installed
+                    const binaryPath = "/usr/bin/openssl";
```
NuSkooler, line 125: No need to check if a binary exists before trying it, just try it. You'll get a not found error or the like. 

I'm wondering if we should just use "openssl" here, since `/usr/bin/` isn't always it's home

core/oputil/oputil_ssh_key.js
```diff
@@ -0,0 +1,158 @@
...
+                    if (fs.existsSync(targetKeyFile)) {
+                        ui.log.write(`${targetKeyFile} already exists.`)
+
+                        return callback('exit');
+                    }
+
+                    ui.log.write(`Creating SSH Key: ${targetKeyFile}`);
+
+                    // Create Dir
+                    if (!fs.existsSync(sshKeyPath)) {
```
NuSkooler, line 119: Please use the fs-extra module, it's already a dep. Same thing here, no need to check if it exists, just try and ignore if it does exist.

https://github.com/jprichardson/node-fs-extra/blob/HEAD/docs/ensureDir-sync.md

core/oputil/oputil_ssh_key.js
```diff
@@ -0,0 +1,158 @@
...
+                ui.log.write("SSH Keys Generated")
+            }
+        }
+    );
+}
+
+function createNew(cb) {
+    const ui = new inq.ui.BottomBar();
+
+    let sslPassword;
```
NuSkooler, line 78: this can just be assigned below, e.g. `const sslPassword = answers.password.trim()`

core/oputil/oputil_ssh_key.js
crhultay: Done

core/oputil/oputil_ssh_key.js
crhultay: Done

core/oputil/oputil_ssh_key.js
crhultay: Done

core/oputil/oputil_ssh_key.js
crhultay: Done

I'm used to always doing checks such as these in my day job, because it is my experience that smart people will spend hours burning away their mental CPU power on trivial things that could have otherwise been made obvious to them.

As an upcoming recommendation, I would like to implement a pre-flight check routine in the BBS Startup. It would:
- Validate all paths, filenames, and binary references in config.hjson
- Check if ports are already in use for content and login servers
- Produce actionable feedback so that a newbie / returning SysOp has a good experience
- Probably some other stuff, but this seems like a good start

NuSkooler: Certainly check the error code, this just eliminates a pre-condition that can succeed while the real create/open may still fail.

That all sounds great! For pre-checks, if you want to tackle it, a JSON schema for the config could do most of the work, see https://github.com/NuSkooler/enigma-bbs/issues/281

core/oputil/oputil_ssh_key.js
```diff
@@ -0,0 +1,139 @@
...
+                    const sshKeyPath = "config/security/";
+                    const sshKeyFilename = "ssh_private_key.pem";
+                    const targetKeyFile = sshKeyPath + sshKeyFilename;
+
+                    ui.log.write(`Creating SSH Key: ${targetKeyFile}`);
+
+                    // Create Dir
+                    if (!fs.pathExists(sshKeyPath)) {
+                        ui.log.write(`Creating Directory: ${sshKeyPath}`);
+                        exec(`mkdir -p ${sshKeyPath}`);
```
NuSkooler, line 110: Sorry, I should have made more clear what I was pointing out: fs.ensureDirSync() for example vs execing() another process. That removes the need for checking if it exists, see previous link.

crhultay: Ah, that's much easier.

I mainly code in C#/C++, so I'm used to breaking everything into steps. We don't have a magical function that checks if exists, creates the dir, and gives you a hug.

NuSkooler (approved):
🤘

PR #580, opened 2024-11-25
Installation Script Enhancement and Startup Script

crhultay:
Added Selection Menu to choose Dependencies, or to Install Everything
Added Python Environment Installation Option
Added Bugfix for Gopher Template Installation which was throwing path errors

**They key motivation is:**
This enhancement will allow the user to auto-install the Python Runtime Environment with a version proven to work with ENiGMA½. It will remove the inability for a user to run the latest Linux Distro and decouples ENiGMA½ from the operating system's Python environment.

It is a common, and poor practice to downgrade the OS Python Environment as it can break operating system functionality.

It is becoming common, and recommended practice to build your own python environment for resiliency in production systems, and this makes ENiGMA½ more versatile and reliable.

### DIFF
--- a/autoexec.sh
+++ b/autoexec.sh
@@ -1,0 +1,26 @@
+#!/bin/bash
+
+# Setup Node Environment
+export NVM_DIR="$HOME/.nvm"
+[ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"  # This loads nvm
+[ -s "$NVM_DIR/bash_completion" ] && \. "$NVM_DIR/bash_completion"  # This loads nvm bash_completion
+nvm use 18
+
+# Setup Python Environment
+export PYENV_ROOT="$HOME/.pyenv"
+[[ -d $PYENV_ROOT/bin ]] && export PATH="$PYENV_ROOT/bin:$PATH"
+eval "$(pyenv init -)"
+pyenv local 3.10
+
+# Start BBS
+/home/egonis/enigma-bbs/main.js
+result=$?
+
+# Determine whether a Startup Crash Occurred
+if [ $result -eq 0 ]; then
+	echo "$result"
+else
+	echo "FAIL: ENiGMA½ exited with $result"
+
+	# TODO: Notify via SMS / Email of Startup Failure
+fi
--- a/autoexec.sh
+++ b/autoexec.sh
@@ -3,6 +3,7 @@
 # Environment
 ENIGMA_INSTALL_DIR=${ENIGMA_INSTALL_DIR:=$HOME/enigma-bbs}
 AUTOEXEC_LOGFILE="$ENIGMA_INSTALL_DIR/logs/autoexec.log"
+TIME_FORMAT=`date "+%Y-%m-%d %H:%M:%S"`
 
 # Mise en place
 ~/.local/bin/mise activate bash >> bash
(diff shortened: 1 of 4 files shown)


autoexec.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+
+# Setup Node Environment
+export NVM_DIR="$HOME/.nvm"
```
NuSkooler, line 4: I've actually been thinking about switching to [mise](https://mise.jdx.dev/) for Node, Python, and hopefully anything else we may need at some point. I've switched to it on a few machines and quite link it. Thoughts?

crhultay: It looks like it would be elegant for both node and python.

crhultay: I played with mise, and am finding it a bit frustrating.

Followed the documentation, but although I am doing:
```
mise install python@3.10
mise use -g python@3.10
mise activate bash
```

I am unable to run 'python' to verify that it is in the path.

In addition, the mise instructions state "to install npm, run: mise use -g node", however this does not result in npm being available in the PATH.

Could you share some samples?

NuSkooler: I haven't run into that personally, but I also haven't tried putting it into a script - did you source the full mise environment?

crhultay: Figured most of it out.

Next problem is that when I run ./autoexec.sh it will work in my shell session but not when executed as a systemctl service (will commit this soon), as I get the error:
```
[2024-11-25T03:27:19.067Z]  WARN: ENiGMA½ BBS/60764 on bbs: Exception during "moduleInitialize" (fullModulePath=/home/egonis/enigma-bbs/core/user_2fa_otp.js)
    error: The module '/home/egonis/enigma-bbs/node_modules/node-pty/build/Release/pty.node'
    was compiled against a different Node.js version using
    NODE_MODULE_VERSION 108. This version of Node.js requires
    NODE_MODULE_VERSION 109. Please try re-compiling or re-installing
    the module (for instance, using `npm rebuild` or `npm install`).
```

Feels like it's using the system node, or something...

NuSkooler: The service unit probably needs to point to the exact node bin `which` is under the mise install area. Might be a better scriptable way

crhultay: Binaries are here:
```
~/.local/share/mise/installs/python/${ENIGMA_PYTHON_VERSION}/bin
~/.local/share/mise/installs/node/${ENIGMA_NODE_VERSION}/bin
```

autoexec.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
```
NuSkooler, line 1: Love the filename 💾 

Maybe `#!/usr/bin/env bash` to match other bash scripts here

crhultay: Done

crhultay:
autoexec.sh confirmed working with systemctl

crhultay:
@NuSkooler I have completed writing for this PR and have tested installation, autoexec, and systemd startup with success on my Ubuntu Server 24.04 installation.

Versions of node and python are now configurable via mise.toml and read automatically into autoexec.sh for pre-startup dependency configuration.

crhultay:
> > @NuSkooler I have completed writing for this PR and have tested installation, autoexec, and systemd startup with success on my Ubuntu Server 24.04 installation.
> > Versions of node and python are now configurable via mise.toml and read automatically into autoexec.sh for pre-startup dependency configuration.
> 
> This looks awesome! Since this is a fairly large change to the initial setup, I'm going to create 1-2 from scratch VM's and run through the paces before merging in (1 Debian/Ubuntu, another RHEL based). Skimming over the code, I don't expect any major issues.

Makes perfect sense to me.

NuSkooler:
First pass with a clean Ubuntu 24.04.1 LTS/Desktop:
Ran the script piped to bash, it just displayed the following and exited.

image

NuSkooler:
> Will evaluate this after some day job stuff, thank you for testing and pardon the issue.

I'll poke around as well, at least it's consistent!

crhultay:
https://stackoverflow.com/questions/46402174/bash-built-in-command-select-not-work-via-pipe-in-shell-script talks about this exact issue

crhultay:
> > Will evaluate this after some day job stuff, thank you for testing and pardon the issue.
> 
> I'll poke around as well, at least it's consistent!

Fixed, try again

crhultay:
Nice catch, fixing

crhultay:
Chicken and the egg issue:
mise.toml isn't in the master branch, so mise doesn't know what to install when we call 'mise install'.

crhultay:
Please approve #582  to remove mise.toml blocker

crhultay:
Mostly done, however time is against me. I will try to have this resolved and working with a scratch install by tonight.

crhultay:
Committed the latest install.sh version; however I get this error during 'npm install':
```
npm error code 127
npm error path /home/egonis/enigma-bbs/node_modules/node-pty
npm error command failed
npm error command sh -c node-gyp rebuild
npm error /usr/bin/env: ‘node’: No such file or directory
npm error A complete log of this run can be found in: /home/egonis/.npm/_logs/2024-11-26T01_07_13_175Z-debug-0.log
```

The issue is that 'node' is an alias, and not found in the PATH. When I run 'which node', nothing is returned.

I'm not entirely sure how to get around this nuance, however it is the final blocker for an install script that works on a vanilla machine.

crhultay:
Spoke too soon -- SORTED!

The solution was to use shims per https://mise.jdx.dev/cli/activate.html

NuSkooler:
I was able to get a bit further, but `mise`, `node`, etc. are not in my PATH even after creating a new terminal session, so enig/oputil are not able to start.

crhultay:
> I was able to get a bit further, but `mise`, `node`, etc. are not in my PATH even after creating a new terminal session, so enig/oputil are not able to start.

If you run autoexec.sh, it will work because it adds the shim to your PATH.

What we could do is echo this into the user's ~/.bashrc and then they would be free to run ./main.js without having to init the mise environment or having to run autoexec.sh.

What do you think?

crhultay:
> npm install -g pm2
This is now fixed, please re-test.

Please note that you must be in the enigma-bbs dir in order for your node/npm/python commands to be functional as the .venv and mise.toml are present in enigma-bbs.

Let me know what you think.

crhultay:
Added some ANSI codes to make it look 31337, I tested this thoroughly in Ubuntu 24.04 with no node installed, and Python 3.12 used by system -- this appears to work great.

NuSkooler:
Just tried out the latest on Ubuntu right before you pushed the ANSI I believe. Worked great! I like that it localizes the install to the enigma directory now 👍 

One note: The first message output in final steps notes `nvm` still.

I'll try Alma next, should pick up your latest ANSI stuff.

autoexec.sh
```diff
@@ -0,0 +1,78 @@
...
+~/.local/bin/mise activate bash >> bash
+
+# Environment Versions
+ENIGMA_NODE_VERSION=${ENIGMA_NODE_VERSION:=$(toml get --toml-path=$ENIGMA_INSTALL_DIR/mise.toml tools.node)}
+ENIGMA_PYTHON_VERSION=${ENIGMA_PYTHON_VERSION:=$(toml get --toml-path=$ENIGMA_INSTALL_DIR/mise.toml tools.python)}
+
+# Validate Environment
+DEPENDENCIES_VALIDATED=1
+
+echo "$(date '+%Y-%m-%d %H:%M:%S') - START:" | tee -a $AUTOEXEC_LOGFILE
```
NuSkooler, line 17: These echo's with dates and such could be turned into a function similar to the other script. You _could_ actually share them between the two by sourcing them into both

NuSkooler:
I don't want to hold this up, so when you're comfortable let me know. It's working great on my end.

crhultay:
autoexec.sh cleaned up to use a common log function, and also auto-detect whether the logs directory exists; turns out it's created on first-run but this will cause an error scenario for new BBSes if ./main.js has not been run the first time.

Please do me a favour and:
```
git add --chmod=+x ./autoexec.sh
```

That is, if this is easy enough for you to do. I am unable to do this in my remote repo without getting fancy.

crhultay:
> Just tried out the latest on Ubuntu right before you pushed the ANSI I believe. Worked great! I like that it localizes the install to the enigma directory now 👍
> 
> One note: The first message output in final steps notes `nvm` still.
> 
> I'll try Alma next, should pick up your latest ANSI stuff.
> 
> Update: AlmaLinux 9 worked like a charm.

Nice catch, replaced 'nvm' with 'mise' and some other touch-ups.

Please note that when you merge this commit, you can probably close out a lot of the issues pertaining to "Enigma isn't compatible with modern distros, etc." since this environment isolation will fix that.

crhultay:
Unrelated note -- my user 'egonis' on your BBS is getting 'That user is already logged in' when I try to connect by telnet, and I get 'Error -21 activating session' via SSH via Syncterm (could be my issue).

Could I get you to look at this? I am missing out on my LoRD character and Dank Dungeons daily routines. :)

NuSkooler:
> Unrelated note -- my user 'egonis' on your BBS is getting 'That user is already logged in' when I try to connect by telnet, and I get 'Error -21 activating session' via SSH via Syncterm (could be my issue).
> 
> Could I get you to look at this? I am missing out on my LoRD character and Dank Dungeons daily routines. :)

Updated - There is a bug in which SSH gets stuck at times :D 

Merging and will apply +x!

NuSkooler (approved):
This is going to help a lot of new users!

NuSkooler:
> Do we have an issue for the SSH issue and how to reproduce?

I could have sworn there was a issue, but I don't see it. Feel free to create. It's something I want to start looking at as well as I get some time here.

PR #581, opened 2024-11-25
Create watch.sh

crhultay:
bash script for newbies which uses bunyan from node_modules to pipe the logfie through bunyan for easy access.

As a JS newbie, I found it confusing and not easy at first. This script will allow a sysop to simply watch their logs in real-time by running this command if they so desire for troubleshooting, or just funsies.


NuSkooler (approved):
Awesome!

PR #590, opened 2025-02-23
NodeJS Upgrade and watch.sh +x Fix

crhultay:
Upgraded to NodeJS 22 and Python 3.12 in mise.toml
Updated packages.json
Added +x bit to watch.sh


NuSkooler:
@crhultay Nice! I'll pull down and do some spot checks. There are some other spots in the docs and what not that will need updated still I believe.

Issue #547, opened 2024-08-17
Starting a second instance on the same host doesn't throw an error or exit


voipmeister:
**Describe the Bug**
Not quite a bug (but no feature either I think), but while checking different versions of Node and Enigma I accidentally started a second instance. The second instance wasn't bound to port 8888/tcp ofcourse, but there was no warning message either. Would be nice to have a warning message and exit IMHO.

**To Reproduce**
Any relevant steps to reproduce the behavior:
Deploy a second installation of Enigma on the same host; start the first instance and the start the second instance.

**Desired Behavior**
Would be nice to have a warning message and exit IMHO.

**Actual Behavior**
The second instance isn't bound to port 8888/tcp and there's no warning message either.

**Environment**
* [x] I am using Node.js v14.x LTS or higher
* [x] `npm install` or `yarn` reports success
* Actual Node.js version (`node --version`): `v18.20.4`
* Operating system (`uname -a` on *nix systems): `Linux bbs01 6.1.21-v8+ #1642 SMP PREEMPT Mon Apr  3 17:24:16 BST 2023 aarch64 GNU/Linux` (Raspberry Pi 3B)
* Revision (`git rev-parse --short HEAD`): `9c0bc88f`
* Any additional relevant information about your setup: None, I think :)

NuSkooler:
@voipmeister There should be errors in your log about binding to ports already in use. Enigma doesn't just panic and exit in this case, however.

voipmeister:
Understood. Would it be possible to provide a little feedback on the console though? Granted, in production you'd not run into this often or easily, but I'm testing with various installs on the same VM and well... You know :)

NuSkooler:
Fixed in #731, on `master` as edcad016.

@voipmeister — you were right, and it was a bit worse than "no warning message". Node's `server.listen(port, host, cb)` callback is a one-shot `'listening'` listener: it never receives an error and never fires at all on a failed bind. Every server here treated it as error-first, so `EADDRINUSE` left the startup sequence waiting on a callback that would never arrive — which is why the second instance neither complained nor exited nor finished starting. Telnet *did* notice the event, but logged it at `info` to the rotating log file, and the default logging config has no console stream, so nothing reached your terminal.

Starting a second instance now gives you:

```
Starting servers...
  ssh       FAILED: *:8889 is already in use - is another ENiGMA½ instance or another service already running?
```

and the process exits `1`.

[...]

PR #592, opened 2025-02-24
Added Startup Server Feedback and Port In Use Detection

crhultay:
**Startup with Server Feedback**
image

**Startup with Port In Use Error**
image

This will address @voipmeister's request per #547.

### DIFF
--- a/core/bbs.js
+++ b/core/bbs.js
@@ -13,6 +13,7 @@ const SysLogKeys = require('./system_log.js');
 const UserLogNames = require('./user_log_name');
 
 //  deps
+const COLOUR_CODES = require('./constants/ansi/colour_codes').COLOUR_CODES;
 const async = require('async');
 const util = require('util');
 const _ = require('lodash');
@@ -53,6 +54,8 @@ function printVersionAndExit() {
 }
 
 function main() {
+    console.info(`${COLOUR_CODES.GREEN}Starting up...${COLOUR_CODES.WHITE}\n`);
+
     let errorDisplayed = false;
 
     async.waterfall(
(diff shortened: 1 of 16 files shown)


NuSkooler:
@crhultay This is great work, thank you! 

One note on the color: Please use the existing pipe code based system already in use in code such as in `connect.js` via `pipeWrite()`

crhultay:
> @crhultay This is great work, thank you!
> 
> One note on the color: Please use the existing pipe code based system already in use in code such as in `connect.js` via `pipeWrite()`

```pipeWrite()``` is an extension of ClientTerminal; however the code I have written is to ```console.{level}```.

Looking inside of client_term is a call to ```renegateToAnsi()```.

Do we need to write a new extension to console.prototypes? (not entirely sure how to do this in JS)

**Comment by NuSkooler** [comment]
@crhultay Sorry, yes just use `renegadeToAnsi()` and `process.stdout()`. Note however, that some may be running this in environments where ANSI is not suitable, or capturing in logs -- thus, you'll want to respect `NO_COLOR` and/or a `--no-color` type CLI option, which makes this more complex.

**Comment by crhultay** [comment]
> @crhultay Sorry, yes just use `renegadeToAnsi()` and `process.stdout()`. Note however, that some may be running this in environments where ANSI is not suitable, or capturing in logs -- thus, you'll want to respect `NO_COLOR` and/or a `--no-color` type CLI option, which makes this more complex.

I'm sorry, but I simply cannot code comfortably in NodeJS.

I spent the past two hours trying to figure out why I'm seeing ```require is not defined``` in a new JS file that I added when I try to import anything.

I will try to return to this at a later date so that I spare everyone my homicidal rage.

**Comment by crhultay** [comment]
Time for a reattempt

**Comment by NuSkooler** [comment]
Closing this in favour of #731, now on `master` as edcad016.

Thank you for this, @crhultay, and sorry it sat as long as it did. You found a real bug that had been quietly breaking startup for years, and the two things you set out to fix were exactly the right ones: a port conflict should say something, and the console should tell the operator what actually came up. Both shipped.

Where #731 differs is the mechanism. The root cause turned out to sit *underneath* the port check — Node's `server.listen(port, host, cb)` registers `cb` as a one-shot `'listening'` listener, so it never receives an error argument and never fires at all on a failed bind. Every server treated it as error-first, which is why `EADDRINUSE` left the startup series waiting on a callback that would never come: no message, no exit, no `System started!`. #731 binds by listening for both `'listening'` and `'error'`, so the failure reaches the caller and the exit path already sitting in `bbs.js` finally runs.

That also let the `tcp-port-used` pre-flight go — a connect probe answers a different question than a bind, and whatever it learns is stale by the time the bind happens. No new dependency.

On the colour question we went back and forth on: that one was mine to answer, and I should have written it rather than sending it back to you. It's in `server_listen.js` now — pipe codes rendered to ANSI only when stdout is a TTY and `NO_COLOR` is unset.

Your startup-feedback screenshots are essentially what shipped:

```
Starting servers...
  telnet    listening on *:8888
  ssh       listening on 0.0.0.0:8889
  http      listening on 0.0.0.0:8080
```

Sorry again for the delay. Please do send more.

crhultay:
Thank you kindly for this, got busy with work and life (and admittedly was pretty frustrated and didn't know what to do next with this PR).

I'm glad this contribution was useful.

NuSkooler:
> Thank you kindly for this, got busy with work and life (and admittedly was pretty frustrated and didn't know what to do next with this PR).
> 
> I'm glad this contribution was useful.

Thank you! Totally get it, I sat on this for a while due to life being busy as well!

PR #604, opened 2025-09-05
Add New "maxConnections" Config Option

viztheman:
Since I'm running my BBS on a tiny one CPU DigitalOcean droplet, I need to limit the amount of active nodes to 1. (For example: running just two Dosemu doors simultaneously will peg the processor completely.)

I dug around in the code and found that there was already a TODO for this item. Since it seemed fairly simple, I went ahead and implemented it. ¯\_(ツ)_/¯ I've been using it on my BBS for a while now, and it seems to hold up pretty well.

Comments/questions/concerns are always welcome. I'm not used to submitting PRs for public projects, so apologies if I've committed any faux pas.

### DIFF
--- a/core/config_default.js
+++ b/core/config_default.js
@@ -15,6 +15,7 @@ module.exports = () => {
 
             menuFile: 'menu.hjson', //  'oputil.js config new' will set this appropriately in config.hjson; may be full path
             achievementFile: 'achievements.hjson',
+            maxConnections: 0	// (<= 0) is Unlimited
         },
 
         term: {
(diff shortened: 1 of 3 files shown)


tracker1:
I would strongly suggest bumping to at least a 2GB droplet, if you are on the 512mb level... better still might be another VPS provider altogether.  The lowest priced entry on DO/Linode isn't even a full thread on a cpu and tends to be over-sold.

robbiew:
This sounds useful. Due to bots, I'm often hitting 72+ nodes. Not sure if that is normal, even at a non-standard port.

viztheman:
I didn't see anything in the documents where this would fit, but I *did* need to add it to the config template. The comment I added to the template should be explanatory enough, I feel.

NuSkooler (approved):

PR #628, opened 2026-02-26
Fix broken and outdated documentation links

SemperFu:
* Fix unsecure, broken, and outdated links in documentation


NuSkooler (approved):
Ty!

PR #637, opened 2026-03-27
Feature/message email notifications queue

Juergen-Coding:
## Summary

This PR adds email notifications for message areas, including a per-area user settings menu inside ENiGMA BBS.

Users can now configure notification preferences for selected message areas and receive email notifications for:

- new posts in an area
- replies to their own posts

The implementation uses a queue-based approach with scheduled processing, so notification delivery is decoupled from message posting.

---

### 1. Queue-based email notification delivery
Instead of sending notification emails directly during message posting, this PR introduces a queue-based workflow:

- notification jobs are created when qualifying messages are posted
- jobs are stored in a dedicated notification queue
- a worker processes queued jobs and sends emails
- throttling is respected between outgoing emails

This keeps message posting responsive and makes delivery more robust.

---

### 2. Scheduled worker integration
The notification worker can be triggered through ENiGMA's event scheduler, which makes the solution suitable for both Linux and Windows environments without requiring a Linux-specific daemon/service model.

The worker:

- polls at configurable intervals
- avoids overlapping runs
- processes the queue until idle
- respects the configured delay between outgoing emails

---

### 3. Per-user notification settings menu
A new **Notifications** entry was added to the Message Menu.

Users can enter the notification settings screen and configure email behavior per eligible message area.

Current per-area options:

- `N` = email on new posts
- `R` = email on replies to the user's own posts

Only areas explicitly allowed by configuration are shown to the user.

---

### 4. Per-area policy support
This PR adds policy handling so that each area can independently allow:

- new topic / new post notifications
- reply-to-own-post notifications

This keeps the feature flexible and prevents notification options from appearing where the sysop does not want them.

---

### 5. Notification UI / ANSI integration
The notification settings screen now uses dedicated art files:

- `NTFMNU.ANS` for the main screen
- `NTFPMPT.ANS` for the prompt area

The dynamically generated area list is rendered inside the menu module, while the static layout remains in ANSI art.

The list was also visually improved so that:

- area titles are highlighted
- conference names are shown in a softer color
- `N:` and `R:` labels are clearer
- `ON` / `OFF` states are color-coded

---

### Core behavior
Notification queueing is integrated into message posting/view logic and related notification policy checks.

### New modules
This PR introduces notification-related modules for:

- queue/database handling
- policy decisions
- queued delivery worker
- notification settings menu

### Menu integration
The Message Menu now includes an `N` command to enter notification settings.

### Configuration
The worker interval and related behavior are configurable through the existing config structure.

---

### New files
- `art/general/NTFMNU.ANS`
- `art/general/NTFPMPT.ANS`
- `config/menus/the_news_bbs-message_base.hjson`
- `mods/notification_db.js`
- `mods/notification_policy.js`
- `mods/notification_queue_worker.js`
- `mods/notification_settings_menu.js`

### Updated files
- `core/message.js`
- `core/msg_area_post_fse.js`
- `core/msg_area_view_fse.js`
- `package.json`
- `package-lock.json`
- `yarn.lock`

---

## Behavior from a user perspective

A user can now:

1. enter the Message Menu
2. press `N` for Notifications
3. view eligible message areas
4. toggle:
   - `1N` for new-post email notifications in area 1
   - `1R` for reply email notifications in area 1
5. quit back to the Message Menu

---

## Testing performed

The feature was tested locally with real message posting and real email delivery scenarios.

Verified during testing:

- queue entries are created correctly
- scheduler-triggered worker runs process queued jobs
- notification emails are sent according to the configured user preferences
- reply notifications only trigger where expected
- new-post notifications only trigger where expected
- the Message Menu entry and notification settings screen work interactively
- the message menu navigation keys still work after returning from the notifications screen

---

## Notes

- This PR intentionally focuses on the functional notification feature and its menu integration.
- Personal/local theme experiments and unrelated ANSI/theme changes were intentionally left out of this PR.
- The worker/scheduler approach was chosen to keep the feature cross-platform and to avoid relying on OS-specific service behavior.

---

## Known limitations / future follow-up ideas

Possible follow-up improvements:

- cleaner documentation for sysops
- additional notification event types
- improved wording/help text in the notification menu
- further ANSI polish for the notification settings screen
- optional batching or digest-style notifications

---

## Why this approach

A queued notification model is safer and cleaner than sending email directly from the posting path.

It provides:

- better responsiveness during posting
- safer retry/throttling behavior
- clearer separation between message creation and email delivery
Auswahl_214
Auswahl_189


core/message.js
```diff
@@ -412,6 +412,14 @@ module.exports = class Message {
             }
         }
 
+        // Edit by Delete and Edit Hack
+        // Hide soft-deleted messages (we mark them by subject = "[DELETED]")
+        // Set filter.includeDeleted = true if you ever want to include them again.
+        if (true !== filter.includeDeleted) {
+            appendWhereClause(`IFNULL(m.subject, '') <> '[DELETED]'`, 'AND');
```
NuSkooler: We don't yet have soft deleted implemented, so this wouldn't work in the master branch

PR #653, opened 2026-04-07
Page +op / Break Into Chat functilnality

NuSkooler:
- Adds page op
- Adds break into chat style +op / user chat, classic style


tmccaff57 (commented):
@NuSkooler  chatInfoFormat10: "|08Chatting with |15{partnerName}|08 \xb7 {duration} \xb7 |07ESC|08 to end|07"

Giving Error: Bad string at line 408,75 >>>                 chat ... in my main menu file

PR #655, opened 2026-04-09
Initial Wide Character support

NuSkooler:
See this discussion: https://github.com/NuSkooler/enigma-bbs/discussions/588


NuSkooler:
This is working quite well and seems to resolve at least most issues with Wide Character support. Pretty fun to use, too. Additional MRs can cover updates/bugs/etc.

I'm going to go ahead and merge in to master.

PR #660, opened 2026-04-12
Implement server-side baud rate emulation

NuSkooler:
Replace the SyncTERM/cterm escape-sequence approach to baud emulation with a collect-and-drip mechanism that runs entirely on the server.

Previously, art.display() wrote all bytes to the socket immediately and sent a proprietary ESC sequence asking the terminal client to throttle its own rendering. This only worked in SyncTERM-compatible clients, and the rate was "sticky" — persisting across menus until another menu explicitly reset it, or the session ended.

Now, when a menu config specifies `baudRate`, art.display() collects all encoded output (literals + ANSI control sequences) into a buffer during the synchronous parse phase, then drips it to the socket at the calculated byte rate (~60fps chunk cadence) via a new ClientTerminal.dripWrite() method. The completion callback fires only after the last byte is sent, keeping MCI view setup correctly sequenced. The rate resets automatically when display ends — no escape sequence, no terminal state to clean up.

- core/client_term.js: add dripWrite(buf, bytesPerSec, cb)
- core/art.js: collect-and-drip path when options.baudRate > 0
- core/theme.js: forward baudRate through displayPreparedArt()
- core/menu_module.js: remove setEmulatedBaudRate() escape sequence from beforeArt(); baud handling is now inside art.display()
- core/ansi_term.js: remove setEmulatedBaudRate(), its export, and the emulationSpeed control code entry (no remaining callers)
- test/baud_emulation.test.js: 12 tests covering dripWrite behaviour and the art.display() baud path
- docs + WHATSNEW.md: update baudRate docs; add modem era reference table; correct beforeArt() description in menu-modules.md

### DIFF
--- a/WHATSNEW.md
+++ b/WHATSNEW.md
@@ -3,7 +3,10 @@ This document attempts to track **major** changes and additions in ENiGMA½. For
 
 ## 0.2.0-beta
 
+* **Server-side baud rate emulation** — `baudRate` in a menu's `config` block now throttles art display on the server rather than delegating to a SyncTERM-specific terminal escape sequence. Emulation now works with every terminal client. The previous approach was sticky (rate persisted across menus until explicitly cleared); the new approach is scoped precisely to each art display and resets automatically. Existing `baudRate` config values require no changes.
+
 * **SQLite driver migrated to `better-sqlite3`** -- This is an internal change with no impact on existing data or configuration. Results in some major DB performance gains.
+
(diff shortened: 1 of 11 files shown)


tracker1:
One minor thing... I would work to ensure that any ansi escape sequences are sent whole... since sending an escape character without at least the [ or next sequence character may result in direct rendering depending on latency or other issues.  Minimally a look ahead and sending escapes with the next character as one character shouldn't affect the rates too much, but reduce the risk of a mis-render.

Another option in practice, that I've done is a slight pause after each line of the file is sent... \n, which still allows for a slower rendered effect, but not quite as much overhead in practice... just a couple thoughts.

PR #663, opened 2026-04-13
Update login.in.hjson

tmccaff57:
Had an extra bracket } bug


tmccaff57:
I have to look into this more

PR #667, opened 2026-04-14
fix: ACS parser bugs, grammar improvements, Peggy migration, and tests

NuSkooler:
Bug fixes in the PEG grammar (misc/acs_parser.pegjs) -----------------------------------------------------
* AC (Achievement Count): used undefined variable `points` instead of `count`, and `isNan` instead of `isNaN` -- always threw, returning false for every achievement count check.
* AP (Achievement Points): same `isNan` capitalization bug.
* AA (Account Age): `accountCreated.diff(moment(), 'days')` returned a negative number (past minus now). Inverted to `now.diff(accountCreated, 'days')` so the result is positive.
* SE (Services Enabled): case label `'nttp'` instead of `'nntp'` -- `SE[nntp]` never matched.
* NP / PC: referenced `UserProps.PostCount` which doesn't exist; the actual export is `UserProps.MessagePostCount`. Post count and post/call ratio checks silently read `undefined` and always failed.
* NR / KR / PC: division by zero when denominator (downloads, download bytes, or login count) is 0. `~~(Infinity)` truncated to 0, giving wrong results. Added zero-guards that yield ratio 0 when the denominator is absent.
* SE: missing `default` case in the service-name switch. Unknown service names (or typos) returned `undefined` from `.every()`. Now explicitly returns `false`.

Grammar improvements
--------------------
* Whitespace tolerance: optional spaces are now allowed around `|`, `&`, `!`, parentheses, and between implicit-AND operands. Leading and trailing whitespace is accepted. Sysops can write `GM[users]` or `GM[users] | NC5` -- both are valid.
* SE service names are now case-insensitive: `SE[HTTP]` and `SE[http]` are equivalent.

Tooling
-------
* Migrated parser generator from PEG.js 0.10.0 (2017, unmaintained) to Peggy 5.1.0 (2026, active community fork). Output is backward- compatible; grammar file unchanged.
* Added `peggy` as a devDependency.
* Added `npm run build:acs` script to regenerate core/acs_parser.js from the grammar in one command.

Tests (test/acs.test.js -- 85 tests, all new)
---------------------------------------------
* Grammar syntax: simple codes, OR/AND/NOT/implicit-AND, grouping, nested expressions, whitespace tolerance (7 tests), edge cases.
* Check function correctness: GM, ID, NC, NP, SC, AA, AC, AP, EC, TH, TW, NN, UP, DL, BU, BD, NR, KR, PC, WD, LC, PV, SE, plus null-user/null-client edge cases and division-by-zero regression tests for all three ratio functions.
* ACS class integration: check() scoping/defaults, parse error handling, hasMessageConfRead, hasFileAreaWrite, hasMenuModuleAccess, getConditionalValue with conditional arrays and fallbacks.

Documentation
-------------
* docs/_docs/configuration/acs.md: rewrote with grammar reference (operators, precedence, arguments, whitespace rules), categorised ACS code tables, defaults table, uploadAcs scope explanation, conditional values section.
* docs/_docs/servers/contentservers/nntp.md: expanded ACS warning into a table of which codes fail over NNTP (client-dependent) vs which work (user-property-based), with design guidance.


tracker1:
@NuSkooler you've been on a tear lately with commits... nice to see.

Issue #546, opened 2024-08-12
Enigma Rest API


biggiecheesetherat:
I am developing a Discord bot to link my BBS' accounts to their Discord account. Is there a API I can use to send messages to users?

NuSkooler:
Hi @biggiecheesetherat 

There currently is not a REST-like API for these types of things. You can certainly write your own web handler (they are pluggable in enig) that could expose a endpoint. 

For sending the message itself, see:
* https://github.com/NuSkooler/enigma-bbs/blob/master/core/user_interrupt_queue.js
* https://github.com/NuSkooler/enigma-bbs/blob/master/core/node_msg.js#L60

Basically, you need to get the data from Discord -> enigma (perhaps the web handler) then just pend a interrupt item to the user(s).

cognitivegears:
I think there may be a way to do this - if you use the ActivityPub branch you should be able to integrate with that like it's any other Fediverse server - not 100% integrated and that ActivityPub branch hasn't been merged yet, but may be helpful.

NuSkooler:
v1 is now in

PR #599, opened 2025-06-07
Basic REST API

chilledtonic:
Our server needed a basic, read-only web view for users unfamiliar with terminal-land. So I've implemented a basic Read-Only Web API. I understand that activity-pub support is incoming, but it will likely be too heavy for our uses. Would this be a welcome PR, or should I fork Enigma?

## PR: Document Message Area Web API

Added complete docs for the Message Area Web API in `core/message_area_web.js`. This covers the RESTful JSON API that gives read-only access to public message areas.

### What's covered

**Four main endpoints:**
- List conferences: `GET /api/v1/message-areas/conferences`
- List areas in conference: `GET /api/v1/message-areas/conferences/:confTag/areas`
- List messages in area: `GET /api/v1/message-areas/areas/:areaTag/messages`
- Get message details: `GET /api/v1/message-areas/messages/:messageUuid`

**Key features:**
- Read-only, no auth required
- Auto-filters private areas
- CORS enabled for browser use
- Pagination (max 200 messages/page)
- Optional reply inclusion
- Configurable area exposure via `exposedConfAreas`
- Proper error handling and HTTP status codes

**Config options:**
```hjson
contentServers: {
    web: {
        messageAreaApi: true
        exposedConfAreas: {
            local: {
                include: [ "*" ]
                exclude: [ "private*" ]
            }
        }
    }
}
```

Includes JSON response formats, usage examples with curl/JavaScript, and implementation details. This enables web/mobile app integration and external system access while keeping things secure.

PR #685, opened 2026-05-11
Official RESTful API v1

NuSkooler:


NuSkooler:
Inspired by: https://github.com/NuSkooler/enigma-bbs/pull/599

NuSkooler:
Resolves https://github.com/NuSkooler/enigma-bbs/issues/546

PR #700, opened 2026-06-03
Add v86 multi-node shared disk support (#698)

NuSkooler:
All concurrent v86_door sessions for the same image path now share a single SharedArrayBuffer, giving every node a live view of the same disk state. A serialized flush queue writes the buffer back to disk after each session exits, preventing concurrent write races at the OS level.

Multi-node games that use file/record locking (e.g. TradeWars 2002) work correctly when SHARE.COM is loaded from runBatch — the game handles concurrent access the same way it would on a real BBS.

Additional fixes and improvements:
- Fix v86 API breakage: serial_set_modem_status replaced with serial_set_carrier_detect / serial_set_data_set_ready / serial_set_clear_to_send
- Fix oputil v86 console: pass imageSab (not imagePath) to worker, flush SAB to disk on exit
- Fix oputil v86 desktop: Save Image now POSTs directly to the server (no browser download dialog); broken internal buffer path removed
- Fix DORINFO graphics field: set to 2 (ANSI color) instead of 1 (IBM high-bit) — required by RBBS-mode doors such as TradeWars 2002
- Add unit tests for SAB registry and flush serialization
- Update local-doors-v86.md: correct shared-state docs, add multi-node TW2002 example, document oputil v86 tools, add ANSI/DORINFO note
- Add WHATSNEW and UPGRADE entries


tracker1:
@NuSkooler very cool, will need to take a look at this.. my own work in progress is a door server inpired from MajorBBS GHost3 door server which relies on DosBox-X.

https://github.com/TheMajorBBS/GHost3

PR #717, opened 2026-08-08
fix(ssh): ignore hardcoded terminal size from stock cryptlib clients

### LINKED ISSUE #414: NetRunner over SSH off by one issues
NetRunner (v2.0 beta 22 on Windows x64 as of this writing) reports it's terminal dimensions incorrectly when over SSH -- specifically as 80x48 when 80x25 is expected.

This causes issues with omitting trailing LF's at `termHeight` when displaying art causing the screen to drop down by one and throw positioning off.

There may be a workaround for this. Will also be reported to g00r00.

andy5995:
Fixes #414 

I walked through some manual/visual testing with Claude and then vibe-coded a fix. But I hope it helps. If nothing else, hopefully you can pinch some useful code from it, or it helps with debugging later.

### DIFF
--- a/WHATSNEW.md
+++ b/WHATSNEW.md
@@ -3,6 +3,8 @@ This document attempts to track **major** changes and additions in ENiGMA½. For
 
 ## 0.5.0-beta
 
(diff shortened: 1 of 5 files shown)


NuSkooler:
Awesome work, this has been a PITA for a long time.

NuSkooler (approved):

andy5995:
> Awesome work, this has been a PITA for a long time.

Ah nice!

PR #725, opened 2026-08-27
fix(bso): repair un-archived EchoMail export and BSO spec compliance

### LINKED ISSUE #722: ftn_bso.js: direct-attach export to a non-archived node produces malformed BSO filename (missing dot), causing silent, permanent mail nondelivery
First of all, thank you for the previous quick fix!

Version: 0.5.1-beta (commit 571934db1b30c552b9482f6c9326d8e7405e34cf)

Summary

When a node has no archiveType configured, EchoMail/NetMail export falls through to moveFilesToOutgoing()'s direct-attach rename path, which produces a BSO flow-file name with the extension concatenated directly onto the basename — no separating dot (e.g. 43792ae5cut instead of 43792ae5.cut). Files created this way are invisible to the native BinkP mailer: core/binkp/bso_spool.js matches outbound files with the anchored regex

`/^([0-9a-f]{8})\.(flo|clo|ilo|hlo|dlo|out|cut|iut|hut|dut)$/i`

which requires the dot. The message sits in the outbound spool indefinitely, is skipped on every scheduled pull and crashmail dispatch, and no error is logged at any log level — the writer's fse.move() succeeds; it just moves the file to the wrong name.

Root cause — core/scanner_tossers/ftn_bso.js, moveFilesToOutgoing(), line 1378:

```
if ('.pk_' === ext.toLowerCase()) {
    const newExt = self.getOutgoingFlowFileExtension(
        exportOpts.destAddress, 'mail', exportType, exportOpts.fileCase
    );  // returns bare extension text, e.g. "cut" — no leading dot,
        // see getOutgoingFlowFileExtension(), line 230

    const newPath = paths.join(
        outgoingDir,
        `${paths.basename(oldPath, ext)}${newExt}`   // <-- missing dot
    );
    fse.move(oldPath, newPath, nextFile);
}
```

getOutgoingFlowFileExtension() intentionally returns the extension without a leading dot — consistent with its other call site in getOutgoingFlowFileName(), which correctly does `${controlFileBaseName}.${ext}` a few lines earlier (line ~299). This call site is missing that same . separator.

Compounding issue — malformed name gets misdecoded as a different destination if manually "fixed"

The basename left over from the .pk_ temp file is a message serial number (getMessageSerialNumber(), used for the original packet name at line 209), not a net/node pair. bso_spool.js's getNodesWithPendingMail() assumes any 8-hex-char BSO basename is NNNNnnnn (4 hex digits net + 4 hex digits node). So simply adding the missing dot back in is not a safe fix — it makes the file match the scan regex, but the reader then decodes the serial number as if it were an address, producing a phantom destination (in our repro: basename 43792ae5 → decoded as net 0x4379 / node 0x2ae5 → logged as 700:17273/10981, "No host configured, skipping"). The message is silently misrouted to a non-existent node instead of just failing to send.

Suggested fix

Insert the missing dot at line 1378:

`${paths.basename(oldPath, ext)}.${newExt}`

This alone fixes the immediate bug, but note the correct fix for a fully-working direct-attach path also requires generating a real net/node-based basename (as getOutgoingFlowFileName() does) rather than reusing the .pk_ temp file's serial-number name — otherwise multiple unbundled messages queued to the same node before one ships would collide on identical filenames.

Reproduction (confirmed via packet capture)

Configure two or more FTN networks with a message area exporting to a node that has no archiveType configured.
Post/toss a message to that area, destined for that node.
Observe the resulting file in mail/ftn_out/<network>/ — correct 8-hex-char serial number, but no dot before the extension (e.g. 43792ae5cut).
tcpdump a subsequent BinkP session to that node: no M_FILE frame is ever sent for it. QSIZE/session completes normally otherwise — no error anywhere in logs at any log level, including trace.
As a manual recovery test, renaming the file to insert the dot (43792ae5.cut) causes the next session to attempt sending it — but to a decoded address of 700:17273/10981 rather than the real destination, confirmed via "[BinkP/Caller] No host configured, skipping" in the log, with the hex math on the basename (4379/2ae5) matching exactly.
Renaming to the correct net/node basename for the intended destination (00640000.cut for 700:100/0) ships successfully — confirmed via M_GOT from the remote hub and file removal from the spool afterward.

Workaround

Setting archiveType (e.g. ZIP, assuming InfoZip's zip/unzip are installed) on the affected node in scannerTossers.ftn_bso.nodes routes export through createArcMailBundle() instead, entirely avoiding the buggy direct-attach rename path. Confirmed working — subsequent EchoMail exports crashmail immediately with no leftover file in the outbound spool.

Impact

Silent, permanent mail non-deliveryfor any message exported to a node without archiveType configured — which is the natural default for a newly-added FTN peer, since archiveType isn't required by any startup validation and the example config in the docs shows it as present but doesn't flag its absence as a problem.

**Environment**
* [X ] I am using Node.js v14.x LTS or higher
* [ X] `npm install` or `yarn` reports success
* Operating system (`uname -a` on *nix systems):`Linux bbs 7.0.14-4-pve #1 SMP PREEMPT_DYNAMIC PMX 7.0.14-4 (2026-07-07T07:27Z) x86_64 GNU/Linux`
* Revision (`git rev-parse --short HEAD`):571934db

NuSkooler:
Fixes #722.

Four defects in the outbound spool shared by `ftn_bso` (the writer) and the native BinkP mailer (the reader). Each caused mail to sit in the spool undelivered with **nothing logged at any level, including `trace`**. All are checked against [FTS-5005.003](http://ftsc.org/docs/fts-5005.003).

## 1. Un-archived EchoMail export (#722)

A node with no `archiveType` exports bare packets rather than ArcMail bundles. `moveFilesToOutgoing()` renamed the temp `.pk_` onto a BSO netmail flow file name **with no separating dot** — `43792ae5cut` — which neither this mailer nor any other could match.

Three further defects in that same branch, beyond the report:

- The basename was a message *serial number*, not the destination's net/node — so restoring the dot alone would have decoded a phantom address (`43792ae5` → `700:17273/10981`, exactly as the reporter observed).
- It stripped the temp extension with a lower-cased `ext` argument, which `paths.basename()` matches case-sensitively — so `fileCase: 'upper'` yielded `43792AE5.PK_CUT`.
- It never emitted `NewOutboundBSO`, so crashmail could not fire even had the name been valid.

Un-archived packets now ship as flow file references, exactly as ArcMail bundles and NetMail already did. That is also what the spec calls for — §3.1 gives a netmail flow file a one-to-one correspondence with its destination, so a node holds exactly one at a time, while a single export routinely produces several packets (`packetTargetByteSize`).

`git log -L` shows the line predates the 2016 Prettier pass: this path has never worked.

## 2. Direct-attach files were not renamed on the wire

§3.1: a netmail flow file *"must be dynamically renamed at the moment of sending to a remote system with a unique name and extension `pkt`"*. We sent it under its `.?ut` name, so the transfer succeeded but the receiving tosser did not recognise a packet — which is why the reporter saw `M_GOT` from the hub with no delivery. Still reachable for `.?ut` files placed in our outbound by a netmail tracker, another mailer, or by hand.

## 3. Filename case

§2 asks that software *"be able to handle both"* cases. `getNodesWithPendingMail()` used a case-insensitive regexp while `getOutboundFilesForNode()` `stat()`ed an exactly lower-case name — so a node with upper-case flow files was reported as pending, dialed every poll cycle, given a `.bsy` lock, and then queued **nothing**. Bites `fileCase: 'upper'` and any outbound inherited from a DOS-era mailer, where upper case is the norm.

## 4. Point addresses

§2 puts a point's files in a `NNNNnnnn.pnt` subdirectory. `ftn_bso` has always written that layout; `bso_spool` had no knowledge of it at all — so point mail was never sent, and polling a point address served it the **boss node's** mail instead.

## Testing

`1472 passing`. 17 new tests:

- `test/binkp_ftn_bso_integration.test.js` — 5 covering the real `exportEchoMailMessagesToUplinks` waterfall (only `exportMessagesByUuid` stubbed, so no message DB needed), including the multi-packet case a `.?ut` could never have carried.
- `test/binkp_bso_spool.test.js` — 12 covering case-insensitive lookup/locking and point scan/ship/lock/reap.

One existing assertion was updated to the new spec-required `.pkt` wire name.

> [!NOTE]
> `ftn_bso.js` captures `config.js`'s `get` at first require, so `_pushTestConfig` cannot rebind it and whichever suite loads the module first freezes `Config()` for the run (the hazard `test/setup.js` documents). The new writer tests set `mod.moduleConfig` directly; without that they pass under `--grep` and fail in a full run.

## Docs

`WHATSNEW.md`, `UPGRADE.md`, and a clause on the `archiveType` row in `bso-import-export.md`.

`UPGRADE.md` carries the only action-required item: stranded files can't be migrated automatically because their names carry no destination address, so it gives `find` commands for both malformed shapes plus a one-liner to read the destination out of the packet header. It also retires `archiveType` as a workaround, and warns sysops carrying points to check their backlog volume before the first poll after upgrade.


NuSkooler:
## Adversarial review pass

Ran an independent adversarial review over this branch. Three findings; one led to a commit, one was mischaracterized, one is accepted as-is.

### Tests that could not fail — fixed in ab881b01

Three of the new spool tests passed unchanged against `master`'s `bso_spool.js`, meaning they could not have caught a regression in the code they were written for:

- the both-cases test referenced *one* packet from both spellings, so the count came out right whichever won
- the point/boss separation test and the point-zero test both passed because the old reader never descended into `.pnt` at all — not because they correctly discriminated

Now: each spelling references its own packet and the winner is asserted; both boss and point have mail queued simultaneously; a real point sits alongside the `.0` entry. Verified by running the suite against `master`'s `bso_spool.js` — failures rise from 10 to 12.

The both-cases test still passes against `master`, and that's correct: `master` resolves *that* input properly too. It's a behavioral pin against a future "prefer uppercase" or "return both" regression. The case `master` genuinely cannot handle — an upper case flow file with no lower case twin — is covered separately by a test that does fail against it.

### Claimed O(N²) polling regression — not a regression

The review reported this as a confirmed measured regression. It isn't. It benchmarked a *simulated* reimplementation of the old lookup rather than actual `master`. Measured against the real thing, one flat outbound directory, sequential `getOutboundFilesForNode` per node:

| nodes | this branch | `master` |
|------:|------------:|---------:|
|   500 |  1,265 ms | 2,463 ms |
| 1,000 |  2,660 ms | 4,919 ms |
| 2,000 |  6,087 ms | 8,846 ms |
| 4,000 | 18,677 ms | 18,145 ms |
| 6,000 | 38,299 ms | 26,604 ms |

Replacing ten targeted `stat()` calls with one `readdir` makes this branch **~2x faster** through the range any real system occupies, with the crossover at roughly 4,000 nodes queued in a single directory. Beyond that the `readdir` grows with directory size and this branch does become slower.

FidoNet has on the order of a couple thousand nodes worldwide, so 4,000+ downlinks in one hub's outbound is not a reachable configuration. I prototyped a per-instance listing cache to flatten the tail anyway and **backed it out**: `caller.js` reuses one `BsoSpool` across an entire poll loop, so cached listings would go stale for `acquireLock`'s cross-case check — trading a hypothetical scaling wart for a real correctness bug. Recording the characteristic here rather than adding a cache whose lifetime isn't scoped to a single logical operation.

### Accepted as-is

- **TOCTOU between `_resolveExistingBsy()` and the `wx` create.** Real, and demonstrated by injecting a write into the window. Only reachable when a differently-cased external mailer shares the spool concurrently; our own `nodeBaseName()` is always lowercase, so self-vs-self stays serialized by the atomic create. Strictly better than `master`, which had no cross-case protection at all.
- **Dual-case coexistence strands the non-lowercase twin.** Confirmed, and identical on `master` — not introduced here.

### Attacked and held

Concurrent `fs.appendFile` to one flow file (200–500 writers, no interleaving or loss on ext4 — and the `NewOutboundBSO` storm is already debounced in `binkp.js`); `uniquePacketName` collisions (`moveFileWithCollisionHandling` renames rather than overwrites on the receive side); the non-deterministic wire name against `disposeMap` (direct-attach entries carry `disposeFn: null` and are skipped, and the lookup is single-invocation per session on every reachable path); point padding and zone fallback; and every FTS-5005.003 quotation and section number, checked verbatim against the spec, including the `readUInt16LE` offsets in the `UPGRADE.md` recovery script against `ftn_mail_packet.js`.

Suite: 1472 passing. Both new test files verified order-independent across repeated runs in both orders.

PR #738, opened 2026-08-28
fix(ftn): file routed NetMail under the uplink's zone, not the recipient's

### LINKED ISSUE #734: Inter-zone/routed NetMail exported to wrong outbound directory, silently never delivered
Thank you for all the other fixes!!

Component: core/scanner_tossers/ftn_bso.js, exportNetMailMessagesToUplinks

Environment: ENiGMA½ 0.5.1-beta, native BinkP mailer, scannerTossers.ftn_bso.netMail.routes configured with a routed uplink for a different zone than the message's ultimate destination.

Summary: When NetMail is routed through an uplink via netMail.routes (e.g. Zone 2 mail routed through a Zone 1 uplink, per standard FidoNet zone-gate practice), the exported packet is filed into the outbound directory matching the destination's zone rather than the uplink's zone. The BinkP caller session to that uplink then finds nothing to send — it logs a clean Session complete, giving no indication anything is wrong, while the message sits undelivered indefinitely.

Steps to reproduce:

Configure a netMail.routes entry routing an out-of-zone address through an in-zone uplink, e.g.:
 ```
  netMail: {
       routes: {
           "*": { address: "1:154/10", network: fidonet }
       }
   }
```
Send a NetMail to an address in a different zone than the route target (e.g. 2:280/464, routed to 1:154/10).
Observe export succeeds with no error.
Observe the BinkP caller session to 1:154/10 (or crashmail dispatch) completes cleanly.
The message is never received by the destination; the exported .pkt/.clo files remain on disk indefinitely.

Root cause:
In exportNetMailMessagesToUplinks (ftn_bso.js:1096), the outbound directory is computed as:

```
exportOpts.outgoingDir = self.getOutgoingEchoMailPacketDir(
    exportOpts.networkName,
    exportOpts.destAddress   // should be exportOpts.routeAddress
);
```

getOutgoingEchoMailPacketDir appends a zone-extension suffix (e.g. outbound.002) whenever the passed address's zone differs from the network's default zone. Here it's passed destAddress — the message's final recipient — rather than routeAddress, the actual uplink node being dialed. For routed cross-zone NetMail, this files the packet under the recipient's zone extension instead of the uplink's zone.

The BinkP caller session, when dialing a specific address, resolves candidate directories via _spoolTargetsForNode() in core/binkp/bso_spool.js (_candidateDirsForZone), which only checks directories tagged with the dialed address's own zone. Since the uplink is dialed at its correct zone, it never looks in the mis-tagged zone-extension directory where the file actually landed — resulting in a session that completes successfully having transferred nothing.

Compounding this: even after manually relocating the misfiled .pkt to the correct directory, delivery still doesn't occur, because the accompanying flow-control file (.clo/.flo/etc., written by flowFileAppendRefs) contains a hardcoded absolute path to the packet's original (wrong) location. _parseFlowFile in bso_spool.js (~line 448) does a literal fsp.stat() on that stored path and silently skips the entry if it doesn't resolve — so the stale reference has to be corrected too, not just the file's location.

Impact: Any NetMail routed cross-zone through an uplink is silently dropped, with logging that actively suggests success (Session complete) rather than failure. No error is surfaced to the sender or sysop.

Suggested fix: Use exportOpts.routeAddress (falling back to destAddress only when unrouted) when computing the outbound directory for NetMail export, so the packet lands in the directory matching the zone of the node actually being dialed.

Confirmed via: direct log tracing and source inspection — misfiled packet located exactly where the zone-substitution predicts (outbound.002/009a000a.pkt, hex-encoded net/node of the route address), two real completed BinkP sessions to the route address that transferred nothing, and confirmed stale absolute path inside the .clo file matching the original mis-filed location.

**Environment**
* [ X] I am using Node.js v14.x LTS or higher
* [ X] `npm install` or `yarn` reports success
* Operating system (`uname -a` on *nix systems): `Linux bbs 7.0.14-4-pve #1 SMP PREEMPT_DYNAMIC PMX 7.0.14-4 (2026-07-07T07:27Z) x86_64 GNU/Linux`
* Revision (`git rev-parse --short HEAD`):669df2ae
* Any additional relevant information about your setup

### PR DESCRIPTION
**Pull request by NuSkooler**
Fixes #734.

## The bug

NetMail routed through an uplink in a different zone than the recipient — the standard FidoNet zone gate, zone 2 mail leaving through a zone 1 hub — was filed in the outbound directory for the **recipient's** zone. A mailer only looks in the zone of the node it is calling, so the session to the uplink completed cleanly having transferred nothing, and the message sat on disk indefinitely.

`exportNetMailMessagesToUplinks()` passed `destAddress` where its sibling steps in the same series already used `routeAddress` — the packet header (`exportNetMailMessagePacket`), the flow file name, and the `NewOutboundBSO` emit. Introduced by afe0c88c (2018-01-20, "NetMail non-HUB fixes"). Before that commit `exportOpts.destAddress` **was** the route address:

```js
exportOpts.destAddress = routeAddress;
...
exportOpts.outgoingDir = self.getOutgoingEchoMailPacketDir(exportOpts.networkName, exportOpts.destAddress);
```

afe0c88c split the two apart — `destAddress` became the real recipient and `routeAddress` the uplink — and left the `outgoingDir` line untouched as context, so its meaning changed underneath it.

Reproduced against the real modules, with the reporter's configuration (`"*" → 1:154/10`, recipient `2:280/464`, local `1:103/705`):

| | outbound dir | dialing `1:154/10` | pending scan |
|---|---|---|---|
| before | `outbound.002` | **0 files** | `2:154/10` |
| after | `outbound` | 1 file | `1:154/10` |

That `2:154/10` is the part the issue did not catch, and it is why nothing surfaced: the pending scan derives a node from *directory zone + flow file basename*, so the misfiled `.clo` reported a node that does not exist. `pollNodes` found no host for it and skipped it at `debug` level, while the crashmail dial to the real uplink found nothing to send. Every visible signal said success.

## The fix

One identifier — the outbound directory now resolves from `routeAddress`. `routeAddress` is `destAddress` when the message is unrouted, so direct NetMail is unchanged (a test covers that direction too).

## Making this shape of failure noisy

Three changes in `core/binkp/bso_spool.js`, because the silence was the real damage:

- A flow entry whose file is missing is **logged** (once per process per reference) rather than `catch { continue }`.
- `flowHasPending()` now resolves references, so a node whose live entries all point at nothing is **no longer reported as pending** — that state used to put the poller in a dial-every-cycle-and-transfer-nothing loop.
- A reference that does not resolve is retried **by basename in the flow file's own directory**. Flow files store absolute paths (FTS-5005.003 §3.1), so recovering mail misfiled before this fix meant hand-editing them; now it is a matter of moving the files. Deliberately scoped to that one directory — a test asserts it will not pick up a same-named file elsewhere in the tree.

Routed NetMail also logs where it was routed, since which `routes{}` pattern applied is not observable from outside.

## Pattern matching: most specific wins

Adjacent footgun, fixed here because the packet-password case is a real hole. Where several `nodes{}` or `netMail.routes{}` patterns matched an address, the one that applied depended on nothing but `config.hjson` ordering — a catch-all shadowed every override below it.

`getMatchScore()` scoring already existed but only `binkp.nodes{}` used it. It is now `Address.findBestPatternMatch()`, backing all three pattern-keyed tables; `findBestNodeMatch()` delegates to it, and the standing `:TODO:` on `getNodeConfigByAddress()` is closed. Both packet-password lookups were open-coding the same `_.find` and now go through it — **a `"21:*"` block written above `"21:1/100"` meant the specific node's `packetPassword` was never the one checked.** Flagged in UPGRADE.md, since someone may find a password they configured only now starts being enforced.

Routes are still consulted before nodes: `"*": { address: "1:154/10" }` genuinely means "route everything here", and hubs do forward AreaFix. With ordering no longer deciding the outcome, a more specific route can now reliably send particular addresses direct — documented in `docs/_docs/messageareas/netmail.md`.

## Also fixed

Passing an already-loaded `Message` to `exportNetMailMessagesToUplinks()` exported an **empty placeholder** — `message` was never assigned from `msgOrUuid`. Latent, since every caller passes UUIDs today, but the regression test needs the documented form.

## Tests

- `test/binkp_ftn_bso_integration.test.js` — routed NetMail export end to end through the real `exportNetMailMessagesToUplinks` series, then handed to `BsoSpool`, so writer and reader are checked against each other rather than against a restatement of either. Covers cross-zone routing, the reverse direction (in-zone recipient routed out of zone), the pending-scan attribution, and route selection order. Verified these fail before the fix and pass after.
- `test/binkp_bso_spool.test.js` — rescued and dangling flow references, including that post-send bookkeeping still marks the line as written, and that the fallback does not reach across directories.
- `test/ftn_address.test.js` (new) — `Address.findBestPatternMatch()` specificity.
- `test/setup.js` gained `general.boardName`: `core/ftn_util.js` captures `config.js`'s getter at require time, so whichever config is live at first require is pinned for the whole run. Without it any test exercising a real FTN export throws from inside an async series and surfaces only as a timeout.

Full suite green — 1761 unit, 8 live. eslint and prettier clean.

## Upgrading

The fix applies to newly exported mail; it does not relocate what is already queued. `UPGRADE.md` has the procedure, which I verified end to end against a reconstructed misfiled spool: move the flow file and its packets into the uplink's outbound directory, leave the flow file's contents alone, and it ships.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -22,6 +22,16 @@ Refer to [Upgrading](./docs/_docs/admin/upgrading.md) for details around this pr
 
 ## 0.5.0-beta to 0.5.1-beta
 
+* **Cross-zone routed NetMail queued before this release will not have been delivered, and needs moving by hand** ([#734](https://github.com/NuSkooler/enigma-bbs/issues/734)). NetMail routed through an uplink in a zone other than the recipient's was filed in the outbound directory for the *recipient's* zone rather than the uplink's, where no mailer would ever find it. The fix applies to newly exported mail only -- it does not relocate what is already queued.
+
+  **Action:** only if you route NetMail across zones. Look for a zone-suffixed outbound directory that should not hold mail for that uplink -- for a `1:154/10` uplink carrying zone 2 mail, that is `outbound.002/009a000a.*` (`009a` = net 154, `000a` = node 10) plus the `.pkt` files its flow file names. Move both the flow file and the packets into the uplink's own outbound directory (`outbound/` when the uplink is in your network's default zone). **The stored paths inside the flow file do not need editing** -- a reference that no longer resolves is now also looked for by name in the flow file's own directory. If you would rather not sort it out, deleting the stray directory's contents loses the affected messages and nothing else.
+
+  Two related changes are worth knowing about even if none of this applies to you: a flow file entry whose file is missing is now logged rather than silently skipped, and a node whose entries all point at missing files is no longer reported as having pending mail -- so it stops being dialled every poll cycle with nothing to send.
+
+* **The most specific `nodes{}` and NetMail `routes{}` pattern now wins, rather than the first one written.** Where only one pattern matched an address nothing changes. Where several matched, the entry that applied used to depend on the order of your `config.hjson`.
+
+  **Action:** review your configuration if it mixes a wildcard with more specific entries that also match. Two cases change behaviour, both towards what the specific entry says: a `"21:*"` node block written above `"21:1/100"` no longer shadows that node's own settings -- including `packetPassword`, which means a packet password you configured and believed to be in force may only now start being enforced -- and a `"*"` NetMail route written above `"21:*"` no longer claims mail the narrower route was meant to take.
+
 * **The `download` file area ACS is now actually enforced.** It previously applied only to the REST API; over telnet and SSH any user who could browse an area could also download from it. If you set a `download` ACS on any file area expecting it to restrict downloads, **it was not doing so until now**.
 
   **Action:** review your `fileBase.areas` ACS blocks. Areas with no `download` ACS are unaffected -- the default (`GM[users]`) matches the `read` default, so nothing changes for them. Areas where you *did* set `download` will now restrict downloads to what you configured, which may be tighter than what users have actually been getting. Note that an entry whose area is no longer in `config.hjson` fails closed and can no longer be downloaded.
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -32,6 +32,17 @@ Refer to [Upgrading](./docs/_docs/admin/upgrading.md) for details around this pr
 
   **Action:** review your configuration if it mixes a wildcard with more specific entries that also match. Two cases change behaviour, both towards what the specific entry says: a `"21:*"` node block written above `"21:1/100"` no longer shadows that node's own settings -- including `packetPassword`, which means a packet password you configured and believed to be in force may only now start being enforced -- and a `"*"` NetMail route written above `"21:*"` no longer claims mail the narrower route was meant to take.
 
+  **The same rule can also relax a packet password, so check this direction too.** A matching entry is used whole; its fields are not merged with those of the wildcard it beat. So where the wildcard carries the password and the specific entry does not:
+
+  ````hjson
+  nodes: {
+      "21:*"     : { packetPassword: MUHPA55 }
+      "21:1/100" : { archiveType: ZIP }        //  no packetPassword
+  }
+  ````
+
+  packets from `21:1/100` used to be password checked by way of the wildcard and now are not, because the specific entry wins and sets no password. If any of your specific node blocks omit a `packetPassword` that a broader entry supplies, add it to them.
+
 * **The `download` file area ACS is now actually enforced.** It previously applied only to the REST API; over telnet and SSH any user who could browse an area could also download from it. If you set a `download` ACS on any file area expecting it to restrict downloads, **it was not doing so until now**.
 
   **Action:** review your `fileBase.areas` ACS blocks. Areas with no `download` ACS are unaffected -- the default (`GM[users]`) matches the `read` default, so nothing changes for them. Areas where you *did* set `download` will now restrict downloads to what you configured, which may be tighter than what users have actually been getting. Note that an entry whose area is no longer in `config.hjson` fails closed and can no longer be downloaded.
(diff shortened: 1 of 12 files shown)


NuSkooler:
Follow-up from a review pass — 1c04f647.

**A flow file reference could resolve to a directory.** `resolveFlowRef()` accepted anything `stat()` succeeded on. Such an entry queues, fails to open at send time (EISDIR), gets `M_SKIP`ped, and so is never marked sent — leaving its node reported pending on every poll cycle forever. That is the same dial-and-transfer-nothing loop the resolution check was added to break, reached by a different shape.

A bare directory reference behaved this way before that check existed too, but the basename fallback widened it. A wholly dangling reference is rescued by its last path segment, so:

```
ref    : <tmp>/gone/missing/..
queued : [ { path: '<tmp>', size: 4096 } ]     <- parent of the outbound dir
pending: [ '1:104/1' ]
```

`..` lands on the parent of the outbound directory — outside the spool — and `.` lands on the outbound directory itself. Both exist, both stat clean, neither is a file. Now guarded with `stat.isFile()`; both cases have tests, verified failing without the guard.

Two smaller things taken with it: `warnedFlowRefs` is bounded (unlike `warnedNetworks` it has no natural ceiling — a downlink that never answers accumulates a uniquely named packet per export), and the non-ENOENT stat warning goes through the same throttle so a persistent EACCES on part of the outbound tree does not re-log every cycle.

**UPGRADE.md covered only one direction of the most-specific-wins change.** A matching `nodes{}` entry is used whole rather than merged with the wildcard it beat, so:

```hjson
nodes: {
    "21:*"     : { packetPassword: MUHPA55 }
    "21:1/100" : { archiveType: ZIP }        //  no packetPassword
}
```

packets from `21:1/100` used to be password checked by way of the wildcard and now are not. That is a relaxation, and the direction actually worth acting on — the note previously described only the tightening case.

1764 unit tests and 8 live tests green; eslint and prettier clean.

PR #742, opened 2026-08-29
fix(ftn): hold a TIC whose file has not arrived yet, instead of rejecting it (#735)

### LINKED ISSUE #735: TIC control file processed before payload arrives in a separate session; no reconciliation, no retry
Component: TIC processing in ftn_bso.js (ticAreas handling)

Summary: When a .tic control file and its announced payload file arrive in separate BinkP sessions (observed with a Fidonet peer's HTick delivery), the .tic is processed immediately on arrival and fails with ENOENT if the payload hasn't landed yet. The control file is archived to reject/ with no retry mechanism and no way to reconcile a later-arriving bare payload against the already-rejected TIC.

Steps to reproduce:

Configure a ticAreas entry for a peer whose TIC announcements can arrive ahead of the payload file (observed: Fidonet peer using HTick, .tic for a full Zone 1 nodelist arriving 15–20 minutes before the actual payload, in a distinct BinkP session).
Receive the .tic control file.
Observe immediate processing attempt and ENOENT rejection, since the payload isn't present yet.
Observe the payload arrives later with no accompanying re-announcement.
The file is never imported; it has to be manually recovered from mail/ftn_in/secure/ and re-scanned via oputil.js fb scan.

Observed severity: for one specific file/peer combination (a full Zone 1 nodelist delivered via a particular peer's HTick setup), this has a 100% failure rate — it has never once succeeded via normal TIC import since the system went live. A second file from the same peer showed an intermittent version of the same race (succeeded once, failed once).

Impact: Legitimate TIC-announced files can be silently and permanently lost from the perspective of the automated pipeline, requiring manual sysop intervention (locating the orphaned payload in the secure inbound directory, copying it to the correct file-base storage path, and forcing a rescan — which itself has a known dupe-detection gotcha requiring fb rm before rescan if a prior failed attempt left stale metadata).

Suggested fix: When a .tic fails validation solely due to a missing payload file (ENOENT on the expected filename), hold the control file for a bounded retry window (checking on subsequent inbound sessions/scans) rather than immediately rejecting it — analogous to how BSO netmail/echomail tossing already tolerates files arriving across separate sessions. Alternatively, when a bare file later arrives in the inbound directory with a name matching a previously rejected .tic, attempt reconciliation against the rejected control file rather than treating it as an orphan.

**Environment**
* [X ] I am using Node.js v14.x LTS or higher
* [X ] `npm install` or `yarn` reports success
* Operating system (`uname -a` on *nix systems):Linux bbs 7.0.14-4-pve #1 SMP PREEMPT_DYNAMIC PMX 7.0.14-4 (2026-07-07T07:27Z) x86_64 GNU/Linux
* Revision (`git rev-parse --short HEAD`):669df2ae

NuSkooler:
Fixes #735.

Two commits: the #735 fix itself, then the fixes for two issues a security review of that work turned up (kept separate for reviewability; both belong here).

## The bug

A `.tic` and the file it announces routinely arrive in **separate mailer sessions**. The reporter's peer, running HTick, announces a full Zone 1 nodelist 15–20 minutes ahead of the payload.

ENiGMA½ processed the `.tic` the moment it landed, could not find the file, archived the announcement to `reject/` and unlinked it. The payload then arrived with nothing left to pair it with and sat in the secure inbound indefinitely. For that peer and that file it failed **every single time**; recovery meant finding the orphan by hand and forcing a rescan.

A `:TODO:` sat on the exact line for years:

```js
//  :TODO: If ENOENT -OR- failed due to CRC mismatch: create a pending state & try again later; the "attached" file may not yet be ready.
```

## The fix

Hold such a TIC and retry it on later import passes. This is the disposition HTick itself uses — `processTic()` logs *"has not been received, waiting"* and returns `TIC_NotRecvd`; `processDir()` then leaves the TIC in the inbound. An import pass already runs the instant a BinkP session delivers files, so the pairing normally completes within seconds of the file landing.

Age is taken from the TIC's own mtime — no new state to keep, and it survives a restart — bounded by the new `tic.holdMaxAgeMs` (default 48h; `0` holds indefinitely).

The reject-and-reconcile alternative suggested in the issue is deliberately **not** taken: never rejecting the announcement in the first place needs no matching of mangled `reject-tic--<ts>-NAME.TIC` filenames and no guess at which of several rejects is authoritative.

## Security fixes (second commit)

**Arbitrary file write via TIC `Lfile` / `Fullname` — pre-existing.** The `File` field was traversal-checked; the long-name fields were not — and it is the *long* name a file is stored under:

```js
localInfo.fileEntry.fileName = ticFileInfo.longFileName;   // Lfile || Fullname || File
const dst = paths.join(areaStorageDir, localInfo.fileEntry.fileName);
```

No `basename()` or scrubber anywhere between. Confirmed against a live instance: a TIC with a valid `File`, correct CRC and `Lfile ../../ESCAPED-PROOF.txt` wrote peer-controlled content two directories above the file base while the file base stayed empty — a route to code execution. Preconditions were a configured node, a mapped `ticAreas` entry, a matching password where one is set, and a correct CRC, so an authenticated-peer attack in practice; the `secureInOnly` enforcement below independently narrows it to the secure inbound by default. All three fields are now validated, and `longFileName` discards an unsafe candidate rather than returning it. Re-running the exploit post-fix: rejected with `Invalid or unsafe Lfile field in TIC`, both files archived, nothing escapes.

**Name claims honoured from TICs we refuse to process — introduced by the first commit.** `getTicClaimedNames()` ran before any `secureInOnly` gate, so every parseable `.tic` contributed its `File` value to the skip set used by the packet and bundle stages. A `.tic` in the *unsecure* inbound is never processed and therefore never removed, so an unauthenticated peer could drop one naming an arbitrary file and permanently shield it — including real inbound mail — from being tossed. The `secureInOnly` decision now lives in one place, `ticProcessingAllowed()`, used by both the reservation pass and the import pass so they cannot drift.

The review separately scrutinised and cleared the `resolveFilePath()` traversal guard, the case-insensitive password change, `secureInOnly` threading, and the `holdMaxAgeMs` mtime expiry.

## Also fixed, all in the same code path

* **`TicFileInfo#filePath` threw a `TypeError`** for a TIC with no `File` field — `paths.join(dirname, undefined)`. Every error path reads that property, so the throw escaped into an fs callback and left `async.eachSeries` hung, stalling the pass until the 5-minute watchdog fired and skipping every remaining TIC *and* the other inbound directory.

* **The announced name was matched case-sensitively**, so a TIC naming `NODELIST.Z34` never found a delivered `nodelist.z34`. This is an independent, sufficient cause of "never once imported". Now resolved case-insensitively against the TIC's own directory, as HTick does with `adaptcase()`. Only basenames are compared and the traversal check is re-applied, so a match can only be a sibling of the TIC.

* **A file still being written could be destroyed.** A payload shorter than the announced `Size`, or failing its CRC, was archived as a reject *and unlinked* — which, with a mailer that writes straight into the inbound (binkd and friends), could delete a transfer in progress. `Size` is now checked before the checksum so a partial is diagnosed as incomplete, and all such conditions are held rather than rejected.

* **`tic.secureInOnly` did nothing.** Documented and defaulted to `true` since TIC support landed, it was never read — so TIC files in the *unsecure* inbound were imported into the file base on the strength of an unauthenticated `From` line alone. Now enforced; such files are left in place, neither imported nor deleted, and the fact is logged once.

* **The TIC password was compared case-sensitively**, unlike HTick's `stricmp()` and unlike our own packet password check, which already upper-cases both sides.

* **The packet and bundle stages consumed TIC payloads.** They run over the same directory first, and a bundle extension is any day-of-week pair plus a character — so an announced `FOO.SA1` collided and was archived as a reject and unlinked while the TIC needing it was being held. Names a TIC has claimed are now excluded, and the bundle pattern is anchored so only a trailing match counts.

* Drive-by: `err.toStrong()` → `err.message` in the temp-directory failure path, and archiving a file that is not there no longer logs a warning.

## Verification

**37 tests** in `test/ftn_tic_import.test.js`. Of the original 32, **26 fail against pre-fix code**; the 6 that pass describe behaviour that was already correct (CRC-32 check value `0xCBF43926`, intact import, exact-case preference, valid SHA-256, missing-password rejection, unknown node/area rejection) and are there as guard rails. Full suite: 1822 passing, verified order-independent in both directions against the other FTN suites.

Unit tests necessarily stub the file-base half of `processSingleTicFile`, so this was also run against an **isolated live instance** (own config dir, db, file base and spool):

| | pass 1 (TIC alone) | pass 2 (payload arrives) | file base |
|---|---|---|---|
| **pre-fix** | TIC → `reject/` | payload stranded in `secInbound/` | *empty* |
| **post-fix** | TIC held, `reject/` empty | both consumed | `nodelist.z34` |

The live run also confirmed a case-mismatched delivery imports, an unsecure-inbound TIC is left untouched with one actionable `warn`, and a hold logs `info` once then drops to `debug`.

## Upgrade notes

No action required. `UPGRADE.md` covers three things worth reading: how to recover files this already stranded (move the payload *and* its `reject-tic--*` announcement back into the secure inbound — re-announcement by the peer is not needed), the `secureInOnly` behaviour change for anyone relying on unsecure-inbound TIC imports, and the traversal fix.

## Not in scope

TIC-announced files are still never forwarded to downlinks — filed separately as #743.


NuSkooler:
## Follow-up: security review findings addressed (4b2f8fa)

A security review of this branch surfaced two issues. Both are now fixed here, in a separate commit so it can be split out or cherry-picked if you'd rather handle the first one via an advisory.

### 1. Arbitrary file write via TIC `Lfile` / `Fullname` — pre-existing

The `File` field was traversal-checked; the long-name fields were not — and it is the *long* name a file is stored under:

```js
localInfo.fileEntry.fileName = ticFileInfo.longFileName;   // Lfile || Fullname || File
const dst = paths.join(areaStorageDir, localInfo.fileEntry.fileName);
```

No `basename()` or scrubber anywhere between. **Confirmed against a live instance**: a TIC with a valid `File`, correct CRC and `Lfile ../../ESCAPED-PROOF.txt` wrote peer-controlled content two directories above the file base, and the file base stayed empty. That is a route to code execution (`authorized_keys`, a cron file, a `.js` the BBS loads).

Preconditions: a configured node, a mapped `ticAreas` entry, a matching password where one is set, and a correct CRC — so an authenticated-peer attack in practice, which in FidoNet means anyone you peer with. The `secureInOnly` enforcement earlier in this branch independently narrows it to the secure inbound by default.

All three fields are now validated up front, and `longFileName` discards an unsafe candidate rather than returning it. Re-running the same exploit post-fix: rejected with `Invalid or unsafe Lfile field in TIC`, both files archived to `reject/`, nothing written outside the file base.

### 2. Name claims honoured from TICs we refuse to process — introduced earlier in this branch

`getTicClaimedNames()` ran before any `secureInOnly` gate, so every parseable `.tic` contributed its `File` value to the skip set used by the packet and bundle stages. A `.tic` in the *unsecure* inbound is never processed and therefore never removed — so an unauthenticated peer could drop one naming an arbitrary file and permanently shield it, including real inbound mail, from being tossed.

My regression, and the review was right to flag it. The `secureInOnly` decision now lives in one place, `ticProcessingAllowed()`, used by both the reservation pass and the import pass so they cannot drift.

### Otherwise clean

The review specifically scrutinised and cleared: the `resolveFilePath()` traversal guard (only `readdir` basenames compared, re-joined with the TIC's own dir, `isFile()` excludes directories — stricter than master); the case-insensitive password change (a missing `Pw` cannot match, since a falsy configured password returns early); `secureInOnly` threading (single call site, no bypass); and `holdMaxAgeMs` mtime expiry (not remotely settable — BinkP never applies the sender's timestamp, and the future-timestamp clamp is correct).

**Now 37 tests** in `test/ftn_tic_import.test.js`; full suite 1822 passing.

Issue #755, opened 2026-09-02
BSO: outbound queued for a downlink that never polls grows without bound


NuSkooler:
The other half of the queue-hygiene discussion under #743; this one is a feature rather than a bug.

A downlink that stops polling accumulates flow file references and generated `.tic` files indefinitely. Nothing expires them, and nothing tells the operator that a node has been sitting on undelivered mail for months. With TIC forwarding (#743) each forwarded file adds a payload reference plus a `.tic` per downlink, so a hub feeding a dead link accrues these steadily.

## Suggested

- An age threshold after which queued outbound for a node is reported, and optionally pruned.
- A diagnostic listing nodes with pending outbound and the age of the oldest entry — useful well beyond this case; it is the question "who is not collecting their mail?", which an operator currently cannot answer without reading the spool by hand.
- Pruning must remove the generated `.tic` alongside its payload reference, or the outbound accumulates orphans.

Interacts with passthrough areas: a transit file must outlive the last pending send, so any expiry policy has to be the thing that decides when it may finally go.

NuSkooler:
Re-verified against `master` while working the file-echo backlog. Two of the three asks here are already built, and the third was deliberately rejected — so what remains is much smaller than the issue as filed.

### A diagnostic listing nodes with pending outbound and the age of the oldest entry — **done**

`oputil bso status`, shipped with #743's queue-hygiene work:

```
Node                    Queued  Missing  Oldest
-----------------------------------------------
1:218/701                    2        1     36s
1:218/702                    1        1       -
```

`bso list <address>` breaks one node down per entry. Worth noting for this issue specifically: a node whose queued entries have *all* gone missing is deliberately not polled, so `status` is the only place it appears at all — which was the "who is not collecting their mail?" question this asked for.

### Pruning must remove the generated `.tic` alongside its payload reference — **split out to #862**

This turned out to be a live defect rather than a missing feature, so it is filed separately and fixed in #863. `scrubFlowFileRefs()` knew the adjacency rule; `oputil bso prune` did not, so pruning a vanished payload left the TIC announcing it queued *and* sendable. The downlink collected an announcement for a file it never received. The rule now lives once in `bso_util` and both paths call it.

### An age threshold after which outbound is reported and optionally pruned — **half done, half deliberately rejected**

Age is reported: the `Oldest` column above, and per-entry in `bso list`.

Automatic pruning is not, and should not be. The reasoning is already in the shipped help text and holds up:

[...]

PR #763, opened 2026-09-04
feat(bso): tell the operator which node is waiting on a file that is gone

### LINKED ISSUE #754: BSO: a deleted or moved file silently starves every downlink still queued for it
Split out of the queue-hygiene work discussed under #743 — this half is a bug, not a feature.

A BSO flow file stores an absolute path. When the file it names is deleted or moved — a sysop tidying the file base, a file area being reorganised, a TIC-forwarded payload removed — `resolveFlowRef` (`core/binkp/bso_spool.js`) warns once and skips the entry. The node keeps the reference, keeps being reported as having pending mail on some paths, and **never receives that file**. Nothing tells the operator which file, or that it will never arrive.

This predates #743 but that feature makes it easier to hit, because a forwarded payload is queued with the `keep` disposition and therefore *stays* in the file base, where a sysop may reasonably delete it.

## Suggested

- Report the dangling reference against the node and file it affects, not just the path, and not only once per process.
- Decide whether a permanently unresolvable entry should be pruned from the flow file after some age rather than skipped forever.
- The related `Replaces` dequeue added in #743 (`dequeueReplacedForDownlinks`) already removes a superseded payload *and* its companion `.tic`; the same helper is most of what a pruner needs.

See also the separate queue-expiry issue for the never-polling-downlink half.

NuSkooler:
Fixes #754.

A flow file stores an **absolute path** to each file queued for a node
([FTS-5005.003](http://ftsc.org/docs/fts-5005.003) §3.1). When that file is deleted or
moved — a file base tidied, an area reorganised, a forwarded TIC payload removed —
`resolveFlowRef` skips the entry and the node never receives it. Nothing removed the
entry, and nothing told the operator which node was affected.

## What was actually wrong

Narrower than the issue title suggests, and worth stating precisely: `flowHasPending`
already prevents the poll loop, and a dangling entry does not block the *other* files in
the same flow file. What was broken:

1. **The warning was unusable.** `{ flowFile, ref }` — a bare absolute path, no node
   address — logged **once per process**. A system that hit it at boot said so once and
   was silent from then on.
2. **The node most worth hearing about was the one that hid.** A node whose live entries
   have all gone missing is deliberately excluded from `getNodesWithPendingMail` — right
   for the poller, which should not dial someone it has nothing to give — and that was
   also the only view of the outbound anyone had. Demonstrated against a real spool:

   ```
   --- getNodesWithPendingMail (what the poller sees) ---
   [ '1:218/701' ]                      <- 1:218/702 hidden entirely

   --- inspectOutbound ---
    1:218/701   pending 5000  good1.pkt
                missing    -  gone.zip
    1:218/702   missing    -  alsogone.zip   <- now visible
   ```
3. **Nothing ever said the file was not coming**, or what to do about it.

## What this does

**The warning** now names the node and the file, says the file will never be delivered
rather than that an entry was skipped, and repeats hourly — this is a standing fault, not
an event. The node label comes from the address where the caller has one, and is decoded
from the flow file name where it does not (the outbound sweep has no address in hand; it
is working out which nodes exist).

**`oputil bso`** gives the operator the view they otherwise get by reading the spool by
hand:

```
oputil.js bso status          # every node with outbound, including ones the poller skips
oputil.js bso list ADDRESS    # every queued entry for one node
oputil.js bso prune ADDRESS   # drop entries whose file is gone
```

```
Node                    Queued  Missing  Oldest
-----------------------------------------------
1:218/701                    2        1     36s
1:218/702                    1        1       -
```

**Pruning is deliberately not automatic** and does nothing without `--yes`. A file on an
unmounted volume and a deleted one are the same `ENOENT` here, and guessing wrong discards
mail nobody asked us to throw away — the exact failure class the last few PRs have been
removing. `status` tells the operator what to check first, and the docs lead with the case
that matters most: if the file can be restored to the path the flow file names, nothing
needs pruning and the next poll ships it.

The outbound walk behind `getNodesWithPendingMail` is now shared with the report, so the
two cannot drift on the point-directory and case rules; they differ only in the judgement
each makes about what it finds.

## Scope

Only the bug half of the queue-hygiene discussion. #755 (expiry for a downlink that never
polls) stays open — though `bso status` answers its "who is not collecting their mail?"
half, which the issue notes an operator currently cannot answer at all.

## Tests

13 new cases in `test/binkp_bso_spool.test.js` covering the report, the prune (dry run by
default, missing entries only, sent markers preserved, flow file removed when nothing
sendable remains, other nodes untouched) and the warning content.

`2114 passing`, verified over two consecutive full runs. Lint and prettier clean.
Documented in `docs/_docs/admin/oputil.md`.

**One honest gap:** the tests assert the warning's content and that separate missing files
each get reported, but not the hourly repeat itself — that would need clock control I would
have to add API surface to expose. The interval is verified by reading, not by test.

Two bugs I introduced and the tests caught, both the same mistake: `BsoSpool` accepts plain
address-shaped objects everywhere, and I twice assumed an `Address` instance —
`pruneMissingRefs` compared with `toString()` (never matched) and `nodeLabel` logged
`[object Object]`. There is now a shared `sameAddress()` helper.

---

## Update: findings from an independent review

An independent review of the first commit found one issue that would have lost mail, plus
several smaller ones. All are fixed in `e1d2156f`; the branch is now 2 commits.

**`pruneMissingRefs` rewrote flow files without the FTS-5005 `.bsy` lock.** This was the
serious one and I had missed it. `_withFlowFileWrite` is an in-process promise chain, and
`oputil` runs as its own process — so it offered no protection at all against the running
BBS, where `ftn_bso`'s `flowFileAppendRefs` and a live session's mark-as-sent both write
under the `.bsy` lock for exactly this reason (#749). A ref queued between prune's read and
its write was dropped, and if prune's stale view showed nothing sendable left it unlinked
the whole flow file, new entry included. Reproduced by the reviewer, and I confirmed the
fix holds a lock against an external holder:

```
external holder took the .bsy lock: true
prune reported removed: 0
flow file still present: true
contents intact: true

after release -> removed: 1 | flow gone: true
```

Prune now takes the lock, and when it cannot it changes nothing and names the busy node
rather than returning an empty result that reads like "nothing to do".

**`missing` conflated "absent" with "could not read".** `resolveFlowRef` treats every
`stat` failure alike, so a file behind a permissions problem, an unmounted volume or an
exhausted descriptor table was reported as missing — and would have been pruned. That file
is still there and still owed to the node. `ENOENT`/`ENOTDIR` is now `MISSING`; anything
else is `UNREADABLE`, is never pruned, and is explained separately to the operator.

**Smaller ones, also fixed:** `getNodesWithPendingMail` now passes the address it already
has to `flowHasPending`, so the poll path (by far the most frequent) warns with the
zone-qualified label instead of losing the shared throttle key to the file-name fallback;
the warning cache evicts aged-out entries before falling back to clearing itself, which on
a spool with many dangling refs turned the hourly throttle into a firehose; and an
unconfigured `paths.outbound` now gives a clear message instead of a raw `TypeError`.

**Test count correction:** the description above says 13 new cases in the first commit.
It was 12. With the three added here — multi-entry prune (which is what makes the
descending-index splice order matter), the busy-lock case, and the unreadable case — the
branch adds 15. `2117 passing`, verified over two consecutive full runs.

**Checked and found fine** by the reviewer, recorded so it is known these were looked at:
`inspectOutbound`'s line parsing agrees exactly with `_parseFlowFile` on prefixes, blanks
and `~`; the `_forEachOutboundFile` refactor preserves `getNodesWithPendingMail`'s dedupe,
point-directory recursion, point-zero-is-boss rule and zero-byte-is-a-poll-flag handling;
`nodeLabel`'s point branch is correct; `sameAddress` genuinely fixes the plain-object
comparison; and the oputil argument handling, exit codes and `--yes` match sibling modules.

**Still not covered by tests:** the hourly repeat interval itself, and the `oputil_bso.js`
CLI layer has no tests of its own (sibling `oputil_fat.js` does have `test/oputil_fat.test.js`,
so that is a real gap rather than the house style). Both verified by hand.

---

## Update: CLI tests and the DRY note (`967616f4`)

**DRY.** `friendlySize` reimplemented `core/string_util.js`'s `formatByteSize` — and stopped
at GB where the shared one goes to YB, so a large enough queue would have printed a nonsense
unit. Now uses it. Same for the local `pad`/`padLeft` against `string_util`'s `pad()`. Output
is byte-identical before and after.

Left alone: `friendlyAge`. The existing idiom is `moment.duration(...).humanize()`, which
gives "a few seconds" — prose for a screen, where this is a column in a table, so it produces
`3d`/`4h`/`12m`. Easy to swap if one age format across the codebase is preferred.

**Command-layer tests.** `test/oputil_bso.test.js`, six cases covering what is only reachable
through the CLI:

| Case | Guards |
|---|---|
| `prune` without `--yes` leaves the flow file byte-identical | a destructive command becoming destructive by default |
| `prune --yes` removes the missing entry, keeps the rest | the write path end to end |
| `status` shows a node whose every entry is missing | the node `getNodesWithPendingMail` deliberately hides — the feature's reason to exist |
| bad address → non-zero exit + usage | the scripting contract |
| unconfigured `paths.outbound` → readable message, no stack trace | the `TypeError` guard |
| `--yes` while another process holds the `.bsy` → says busy, changes nothing | the bug the review caught, at the layer an operator uses |

They run oputil as a real process: `argv` is bound from `process.argv` when `oputil_common` is
first required, so driving the handlers in-process would mean rebuilding the module tree per
case with a doctored argv — more machinery than the thing under test. `server_listen.test.js`
already shells out this way. ~4s for the six.

Each was checked by mutation rather than assumed: forcing the `--yes` gate open, dropping the
`paths.outbound` guard, and defeating the `.bsy` lock each break exactly one case and nothing
else.

**Correction to an earlier note in this description:** I said the missing CLI tests were "a
real gap rather than the house style" because `oputil_fat.js` has `test/oputil_fat.test.js`.
That was wrong — I repeated it from the review without checking. That file tests
`fat_image.js` beneath the CLI and never touches `handleFatCommand`, argv or output; no
oputil command layer in this repo had tests before this. These are new ground, not a gap
being closed.

Also has `prune` honour `scannerTossers.ftn_bso.flowLockTimeoutMs` — the same knob `ftn_bso`
waits on — rather than always sitting out the 5s default.

`2123 passing`, two consecutive full runs. Suite goes 35s → 39s.

---

## Update: the warning interval is now configurable, and tested (`6cd8e2fa`)

The one behaviour left untested was the warning throttle — and it is the one worth locking,
because a throttle stuck shut is precisely the silence this branch exists to remove.

`flowRefWarnRepeatMs` now sits on the spool alongside `staleLockMaxAgeMs`, reached the same
way through `buildSpool`, defaulting to the hour it already used. Two cases: quiet across
three passes inside the window, speaks up again once it has passed. Both fail if the
throttle is made permanent (the old once-per-process bug) or removed entirely.

It also earns its place independently — a sysop who already knows about a dangling
reference and has not dealt with it yet can now turn it down. Documented under
`binkp.flowRefWarnRepeatMs`.

**The approach I rejected, since it is the obvious one:** stubbing `Date.now` in the test.
Less code, no config, and I had it working. But `Date.now` is process-wide — everything else
in the run sees the shifted clock for the duration, and `.bsy` staleness is clock-based, so
that coupling would have been lying in wait for whoever wrote the next clock-sensitive test.
The config route touches no global.

Not tested, deliberately: the 1024-entry cap eviction. Reaching it needs more than 1024
*distinct* dangling references on one system, and a regression there produces extra log
lines rather than silence.

`2126 passing`, two consecutive full runs.

### DIFF
--- a/core/oputil/oputil_bso.js
+++ b/core/oputil/oputil_bso.js
@@ -40,6 +40,14 @@ function makeSpool() {
     const config = Config();
     const bsoConfig = config.scannerTossers?.ftn_bso || {};
 
+    if (!bsoConfig.paths?.outbound) {
+        //  Without this every command dies in the directory walk with a bare
+        //  TypeError, which says nothing about what is actually missing.
+        throw new Error(
+            'scannerTossers.ftn_bso.paths.outbound is not configured; there is no outbound spool to inspect'
+        );
+    }
+
     return new BsoSpool({
         paths: bsoConfig.paths || {},
         networks: config.messageNetworks?.ftn?.networks || {},
@@ -157,10 +165,12 @@ function cmdStatus() {
         console.info('-'.repeat(47));
 
         let missingTotal = 0;
+        let unreadableTotal = 0;
         for (const node of interesting) {
             const live = liveEntries(node);
             const missing = live.filter(e => 'missing' === e.status);
             missingTotal += missing.length;
+            unreadableTotal += live.filter(e => 'unreadable' === e.status).length;
 
             console.info(
                 `${pad(node.address.toString(), 22)}${padLeft(
@@ -185,6 +195,18 @@ function cmdStatus() {
             );
             console.info('satisfied the files are really gone and not merely offline —');
             console.info('"oputil bso prune <address> --yes" to drop them.');
+        }
+
+        if (unreadableTotal > 0) {
+            console.info('');
+            console.info(
+                `${unreadableTotal} queued ${
+                    1 === unreadableTotal ? 'entry could' : 'entries could'
+                } not be read — a permissions problem, or a`
+            );
+            console.info(
+                'volume that is not mounted. Those are never pruned; see "bso list".'
+            );
         }
     });
 }
@@ -211,13 +233,29 @@ function cmdList(addressArg) {
 
         console.info(`${addr.toString()}`);
         for (const entry of live) {
-            const flag = 'missing' === entry.status ? 'MISSING' : 'ok';
+            const flag =
+                {
+                    missing: 'MISSING',
+                    unreadable: 'UNREADABLE',
+                }[entry.status] || 'ok';
             console.info(
                 `  ${pad(flag, 9)}${padLeft(friendlySize(entry.size), 10)}${padLeft(
                     friendlyAge(entry.timestamp),
                     8
                 )}  ${entry.path}`
             );
+        }
+
+        const unreadable = live.filter(e => 'unreadable' === e.status);
+        if (unreadable.length > 0) {
+            console.info('');
+            console.info(
+                `${unreadable.length} could not be read at all — a permissions problem, or`
+            );
+            console.info(
+                'a volume that is not mounted. Those are not pruned: the file may well'
+            );
+            console.info('still be there. Fix the access and they will send.');
         }
 
         const missing = live.filter(e => 'missing' === e.status);
@@ -249,7 +287,21 @@ function cmdPrune(addressArg) {
     const write = true === argv.yes;
 
     withSpool(async spool => {
-        const removed = await spool.pruneMissingRefs(addr, { dryRun: !write });
+        const { removed, busy } = await spool.pruneMissingRefs(addr, {
+            dryRun: !write,
+        });
+
+        if (busy.length > 0) {
+            //  Not the same as "nothing to do": the entries are still there,
+            //  we simply were not allowed to touch the file.
+            console.info(
+                `${addr.toString()} is busy — a mail session or the tosser holds its`
+            );
+            console.info('lock. Nothing was changed; try again in a moment.');
+            if (0 === removed.length) {
+                return;
+            }
+        }
 
         if (0 === removed.length) {
             console.info(`Nothing to prune for ${addr.toString()}.`);
--- a/core/oputil/oputil_bso.js
+++ b/core/oputil/oputil_bso.js
@@ -1,0 +1,297 @@
+/* jslint node: true */
+/* eslint-disable no-console */
+'use strict';
+
+/**
+ * oputil_bso.js — BSO outbound spool inspection
+ *
+ * Answers the question an operator otherwise has to answer by reading the
+ * spool by hand: who is not collecting their mail, how long has it been
+ * sitting there, and is any of it queued against a file that no longer
+ * exists?
+ *
+ * A flow file stores an absolute path (FTS-5005.003 §3.1). When the file it
+ * names is deleted or moved -- a file base tidied, an area reorganised, a
+ * forwarded TIC payload removed -- the entry can never be sent. The spool
+ * skips it and says so periodically, but nothing removes it, and nothing but
+ * this told the operator which node was affected.
+ *
+ * Commands:
+ *   bso status                 Every node with outbound, and what is wrong
+ *   bso list <address>         Every queued entry for one node
+ *   bso prune <address>        Drop entries whose file is gone (--yes to write)
+ */
+
+const {
+    printUsageAndSetExitCode,
+    argv,
+    ExitCodes,
+    initConfigAndDatabases,
+} = require('./oputil_common.js');
+const { getHelpFor } = require('./oputil_help.js');
+
+exports.handleBsoCommand = handleBsoCommand;
+
+// ─── helpers ─────────────────────────────────────────────────────────────────
+
+function makeSpool() {
+    const Config = require('../config.js').get;
+    const { BsoSpool } = require('../binkp/bso_spool.js');
+    const config = Config();
+    const bsoConfig = config.scannerTossers?.ftn_bso || {};
+
+    return new BsoSpool({
+        paths: bsoConfig.paths || {},
+        networks: config.messageNetworks?.ftn?.networks || {},
+        defaultNetwork: bsoConfig.defaultNetwork,
+    });
+}
+
+function parseAddress(s) {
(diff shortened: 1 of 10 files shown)


core/oputil/oputil_bso.js
```diff
@@ -0,0 +1,349 @@
...
+    const addr = Address.fromString(s);
+    //  fromString hands back an Address either way; a string it could not
+    //  parse leaves the required parts undefined.
+    if (!addr || undefined === addr.net || undefined === addr.node) {
+        return null;
+    }
+    return addr;
+}
+
+function friendlySize(bytes) {
```
NuSkooler, line 72: We have an existing util for this. DRY

NuSkooler: Fixed in 967616f4 — `friendlySize` now calls `formatByteSize()` from `core/string_util.js`.

You were right and it was worse than a duplicate: mine stopped at GB where the shared one goes to YB, so a large enough queue would have printed a nonsense unit.

Same treatment for two others while I was in here — the local `pad`/`padLeft` were reimplementing `string_util`'s `pad()`, so those now use it too. Output is byte-identical either way; verified against the same fixture before and after.

One I deliberately left alone: `friendlyAge`. The existing idiom is `moment.duration(...).humanize()`, which gives "a few seconds" — that is prose for a screen, and this is a column in a table, so it produces "3d" / "4h" / "12m" instead. Happy to swap it if you would rather have one age format across the codebase.

PR #768, opened 2026-09-05
feat(config): check names that must line up across configuration sections

NuSkooler:
**Part 5/8 of #281.** Stacked on #767 (→ #766 → #765 → #764).

Type checking would not have caught a single one of the expensive configuration bugs in this project's history. The couplings that actually break are **referential** — and none of them produce an error today. The file quietly never lands, the echo is quietly never exported, and nothing in the log says why. `core/bso_util.js` exists in large part because of one of these.

```
config.hjson: 2 issues (2 errors)

  error    fileBase.areas.nodelists.storageTags[0]
           storage tag "nodelist" is not defined in fileBase.storageTags -- did you mean "nodelists"?

  error    scannerTossers.ftn_bso.ticAreas.fidonet_nodelist.network
           FTN network "fidonet" is not defined in messageNetworks.ftn.networks
           known: agoranet
```

## The eight rules

All answerable from the merged configuration alone:

| From | Must name |
|---|---|
| `fileBase.areas.<t>.storageTags[]` | `fileBase.storageTags` |
| `ticAreas.<t>.areaTag` | `fileBase.areas` |
| `ticAreas.<t>.storageTag` | `fileBase.storageTags` |
| `ticAreas.<t>.network` | `messageNetworks.ftn.networks` |
| `messageNetworks.ftn.areas.<t>.network` | same |
| `netMail.routes.<p>.network` | same |
| `nntp.publicMessageConferences.<c>` | `messageConferences` |
| `nntp.publicMessageConferences.<c>[]` | that conference's areas |

The first three are the coupling [binkp.md:272-285](docs/_docs/messageareas/binkp.md#L272-L285) spells out for sysops to keep consistent by hand. `validateOutboundConfig()` already covers `defaultNetwork` and the spool's own ambiguities, so those are **translated rather than reimplemented**; two of its three codes have no equivalent in the general taxonomy and pass through under their own names.

## Suggesting the right thing

An unresolved name is met with the nearest thing the operator **actually configured**, reusing the same routine as unknown keys — these are near misses far more often than wild guesses.

Candidates exclude the entries ENiGMA ships itself (`sys_msg_attach`, `system_internal` and friends). Nobody means to point a file echo at `sys_temp_download`, so offering it is noise. They stay perfectly legal to reference — the shipped areas reference them themselves — and the set is **derived from `config_default.js` rather than listed**, so a new internal area cannot leave a stale list behind. A board with none of its own entries falls back to seeing the built-ins rather than an empty list.

Where there is no near miss, what *is* defined is listed instead, capped at ten so a board with many areas does not turn one finding into a wall of text.

## Avoiding false positives

Three details taken from the code that does the real lookups, each of which would otherwise fire on a correct configuration:

- **Network names match without regard to case** — `canonicalNetworkName()` lowercases both sides, so `AgoraNet` and `agoranet` are the same network. Area and storage tags are matched **exactly**, since their lookups index straight into the object.
- **A `ticAreas` entry may be a bare string**, shorthand for `{ areaTag: <it> }` ([ftn_bso.js:2822-2830](core/scanner_tossers/ftn_bso.js#L2822-L2830)).
- **`storageTags` may be a bare string** rather than an array ([file_base_area.js:247-249](core/file_base_area.js#L247-L249)).

A name that is missing or not a string is left alone — that is a shape problem the type walk already reports, and saying it twice helps nobody.

## Tests

21 new: a passing and a failing fixture per rule, plus a sound configuration and the shipped defaults producing nothing, an empty config not falling over, both shorthand shapes, the case-insensitivity rule, and the three suggestion behaviours (near miss, internals filtered, fallback when the operator has no entries of their own).

Full suite: 2220 passing.


NuSkooler:
Superseded by a replacement PR against `master`. GitHub refuses base-branch changes on a stacked PR, and its base `281/4-startup-surfacing` has now been merged, so this one cannot be retargeted in place. Same branch, same commit, rebased onto `master`.

Issue #133, opened 2017-10-27
Admin/SysOp configuration Interface


danger89:
**Short problem description**
I like to have some simple admin configuration interface. And/or create a default configuration file with **all features** from start (--new).

**Expected behavior**
See `mystic -cfg`, which allows you to change services, settings, ports, user settings, message boards, everything. Just via ANSI/ASCII interface.

**Actual behavior**
Nothing. No configuration area/interface.

NuSkooler:
There is no UI for configuration yet, but have you explored `oputil.js`? There is configuration for quite a few things... the rest is as you say: missing & requires direct editing of HJSON files.

danger89:
Oputil.js - - new generated a too basic configuration file imo. Why not dump all options to hjson file? (even default values)

NuSkooler:
@danger89 I like the idea - I'll look into that. If I'm to go down that path, I'll probably add a `refresh` option as well to add missing values.

There are some that don't even exist in `config.js` as the default is a missing value. Perhaps I'll add them commented out or something.

Please try out some of the other `oputil.js` features as well. There are other options such as importing files, managing areas, etc.

danger89:
Thanks in advance. I will checkout oputil.js some more to find all the features it has so far.. thx

danger89:
I didn't find anything useful in `oputil.js` regarding user management to make users sysop's. I manually need to edit sqlite3?

NuSkooler:
Update on this:
I plan on having `oputil.js config new` create a fully populated `config.hjson` including comments showing example sections and the like. 

An actual configuration UI may still come down the line as well.

NuSkooler:
Update:
`oputil.js config new` will now:

- Create a much more robust `config.hjson` with comments, examples, etc.
- Create a copy of `menu.hjson` and `prompt.hjson` based on your board name

NuSkooler:
@SemperFu started some work on something like this: https://github.com/SemperFu/ENiGMA.Config

danger89:
> @SemperFu started some work on something like this: https://github.com/SemperFu/ENiGMA.Config

That looks fancy indeed.

danger89:
I will close it for now. I guess.

NuSkooler:
@danger89  Reopening as others have asked. It might happen someday, but isn't currently a priority.

crhultay:
I would like to take this issue as my first major contribution.

PR #774, opened 2026-09-05
feat(config): publish a JSON Schema for config.hjson

NuSkooler:
Part of #281. Based on `master` rather than stacked — the six merged PRs showed GitHub's native stacks can't be restacked after a squash-merge.

## What this adds

`misc/config.schema.json` — a [JSON Schema](https://json-schema.org/) (draft 2020-12) for `config.hjson`, generated from the same schema the validator uses by `npm run build:schema`. 84KB, 2069 lines: 76 closed objects, 10 open maps.

It exists for the configuration editor asked for in #133, and for anything else that wants to know what a setting is called, what type it takes and what it defaults to without reading `config_default.js`.

The emitter lives in `core/config/json_schema.js`; `dev_util/gen_config_schema.js` is a thin CLI that only decides where the output goes. Putting the logic in `core/` keeps the projection unit-testable and lets a future editor ask ENiGMA for its schema without a build step. The CLI also takes `--check` for anyone who wants the staleness guard outside of mocha.

## It is a projection, not a second description

Where the artifact and the validator could disagree about a configuration, that is a bug. So:

- **No key patterns for open maps**, which the design floated. The validator does not check the *format* of an open map's keys, so emitting a `patternProperties` constraint would make the artifact reject configurations ENiGMA itself accepts. If key patterns are wanted, they belong in the validator first and this follows for free.
- **`patternProperties: {"^_": true}` on every closed object.** The validator skips `_`-prefixed scratch blocks — `_snips` holding fragments for `@reference` to point at — at every level, so the artifact has to as well.
- **`additionalProperties: false` only where the schema claims to know every key.** An exemplar-derived shape never gets it; that is what keeps the artifact honest about file areas carrying `acs` and `hashTags`.
- **Nothing is ever `required`.** Every setting has a default, and a configuration file may be a fragment completed by `includes`.

Two things a consumer has to know are stated in the emitted document's own `$comment`: it describes the configuration *after* `@reference:` / `@environment:` / `@file:` specs are resolved, and an unrecognised key is an error to a JSON Schema validator but only a warning to ENiGMA, which boots regardless.

## Guards

All mocha tests rather than a workflow step, since CI runs `npm test`.

- **Staleness** — the checked-in artifact must be byte-identical to what the generator produces.
- **Fidelity** — the artifact is walked *as data* by a small checker and must accept both `config_default.js` and the configuration `oputil.js config new` actually writes. That fixture is built the way oputil builds it, importing the now-exported `ConfigIncludeKeys` rather than copying the list: the raw template is not a configuration, since its values are `XXXXX` placeholders and every port in it is a string.
  A **keyword-coverage assertion** fails if the artifact ever grows a keyword the checker does not implement, so the checker cannot quietly become a no-op that passes everything.
- **Annotation completeness** (guard 3 from the design, previously unimplemented) — the set of `meta.js` paths that exist in `config_default.js` is pinned. This is the only guard that catches a *rename*: a renamed `fileBase.areas` would otherwise be grafted straight back on as an open map while the real block sat beside it as a closed object, reporting every area tag a sysop wrote as an unknown key.

## A false positive in the shipped validator, found by making it check itself

The defaults did not validate against their own derived schema.

Exemplar-derived shapes were left open only at the **outermost** level. The archivers in `config_default.js` disagree about the keys of their own `list{}` blocks — only `Arj` carries `entryGroupOrder` — and `mergeShapes` took the first example's block wholesale, closed. So a sysop defining their own archiver was told `entryGroupOrder` was an unknown key, and likewise `processIACs` inside a custom file transfer protocol's `external{}`.

Reproduced against the shipped validator before fixing:

```
archives.archivers.MyArj.list.entryGroupOrder      unknownKey
fileTransferProtocols.myz.external.processIACs     unknownKey
```

`mergeShapes` now recurses: it unions the examples at every level and leaves every level open, and an open map nested inside an example stays an open map rather than having one board's area tags read as a key list. Annotations from `meta.js` survive the union; exemplar `default`s deliberately do not, since a value taken from one example entry is that entry's data.

Verified the real detections still fire — `ticAreas` `hashTag`/`storageTags`, `general.boardname`, top-level `loginServer` — and that `acs` on a file area, a nested archiver key and a conference area stay silent.

## Testing

`npm test` — **2274 passing** (up from 2250), live suite included. `npm run lint` clean.

Not covered: no docs page mentions `oputil.js config validate` or the artifact yet, and I have not re-run this against a large production configuration — that is worth doing before merge given the `mergeShapes` change.


NuSkooler:
## Follow-up: the artifact was not reproducible

Testing on a live board surfaced this immediately, and CI's `test` job failed on the first run for the same reason.

Roughly forty defaults in `config_default.js` are built from `__dirname`, so the generated artifact quoted whatever directory it was generated in:

```json
"areaStoragePrefix": {
    "type": "string",
    "default": "/home/bryan/dev/enigma-bbs/file_base/"
}
```

Two problems. The staleness guard compares the checked-in file to what the *current* checkout produces, so it read the difference between two checkouts as a stale artifact — regenerating on the Xibalba box wrote a different file (84062 vs 84306 bytes) purely because of the path. And a published schema that names one developer's home directory is misleading to anything reading it as a default.

Installation-derived defaults are now described rather than quoted:

```json
"areaStoragePrefix": {
    "type": "string",
    "$comment": "Default: <installation directory>/file_base/"
}
```

It goes in `$comment` rather than `default` deliberately — a consumer that wrote the relative path back into `config.hjson` would be wrong, since the real value is absolute.

Verified byte-identical output from a checkout at an unrelated path, and guarded by a test asserting the artifact contains no reference to the installation root at all. That gives a clear diagnosis instead of the staleness guard's bare "this file differs".

Also confirmed on Xibalba: `config validate` still reports **no problems found**, so the `mergeShapes` change did not regress a real production configuration.

2276 passing, lint clean.

PR #781, opened 2026-09-06
fix(doors): write the DOOR32.SYS comm fields the door is actually given

andy5995:
> direction.

getDoor32Buffer wrote comm type 2 (telnet) with handle -1 whatever the door was actually given, which the code carried as a TODO. Under abracadabra's default that tells a door it has a socket and then gives it no way to reach one, so a door that validates the pair refuses to start rather than read the wrong stream.

DropFile now takes a commType of its own -- local, serial or socket. It is deliberately not Door's io type, which is only ever stdio or socket: abracadabra maps its io onto it, and v86 reports serial, since it bridges the user to the guest's emulated COM1.

Found with [Immortal Barons](https://github.com/andy5995/immortal-barons), which checks the pair: it exited with "line 2 (socket handle): socket mode needs a handle, got -1" and reaches its menus once the dropfile reports local.

### DIFF
--- a/core/abracadabra.js
+++ b/core/abracadabra.js
@@ -156,6 +156,8 @@ exports.getModule = class AbracadabraModule extends MenuModule {
 
                     self.dropFile = new DropFile(self.client, {
                         fileType: self.config.dropFileType,
+                        commType:
+                            'socket' === (self.config.io || 'stdio') ? 'socket' : 'local',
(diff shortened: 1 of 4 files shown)


andy5995:
Superseded by https://github.com/NuSkooler/enigma-bbs/pull/787

PR #782, opened 2026-09-06
fix(views): keep the mask edit cursor aligned after the field is cleared

andy5995:


test/views.test.js
```diff
@@ -526,6 +526,134 @@ describe('MaskEditTextView', () => {
...
+            const h = makeTrackingView('####/##/##');
+            h.view.setFocus(true);
+            h.view.clearText();
+            h.view.setFocus(true);
+
+            '1990'.split('').forEach(ch => h.view.onKeyPress(ch, null));
+            h.view.onKeyPress(null, { name: 'backspace' });
+
+            assert.equal(h.view.lineBuffer.lines[0].chars, '199');
+            assert.equal(h.cursor.col, COL + 3);
```
andy5995, line 653: A second backspace exercises the other arm of the branch — the cursor lands on an input slot rather than on the `/`, so no walk-back happens. Values checked against the branch:

```suggestion
            assert.equal(h.view.lineBuffer.lines[0].chars, '199');
            assert.equal(h.cursor.col, COL + 3);

            //  The second backspace takes the other arm: the cursor is on an
            //  input slot rather than on the '/' literal.
            h.view.onKeyPress(null, { name: 'backspace' });

            assert.equal(h.view.lineBuffer.lines[0].chars, '19');
            assert.equal(h.cursor.col, COL + 2);
```

andy5995 (commented):
> Written by Claude (Opus 5), an LLM made by Anthropic, at andy5995's direction.

I ran the branch against a screen-tracking harness like the one the new tests build.

### The backspace branch still hand-rolls the walk `_patternPosForLength` replaced

Typing, `clearText()` and `setText()` all derive the position from the buffer length now. `onKeyPress`'s backspace arm (`core/mask_edit_text_view.js:211-245`) keeps the decrement, the `assert`, and the nested `while` that walks back over literals — the one site left doing the bookkeeping by hand. That matters because of the TODO this removes:

> There exists some sort of condition that allows pattern position to get out of sync

The code most able to produce that condition is the code that stayed.

The whole branch reduces to the helper:

```js
if (this.isKeyMapped('backspace', key.name)) {
    const textLen = this.lineBuffer.lines[0].chars.length;
    if (textLen > 0) {
        this.lineBuffer.deleteChar(0, textLen - 1);
        this._syncFromBuffer();
        this.patternArrayPos = this._patternPosForLength(textLen - 1);
        this.client.term.write(
            ansi.goto(this.position.row, this.getEndOfTextColumn() + 1)
        );
        this.clientBackspace();
    }

    return;
}
```

The two arms differ only in that the literal case issues an `ansi.goto` first. In the input-slot case that goto resolves to the column the cursor already occupies, so issuing it unconditionally changes nothing — I measured that rather than reasoning it. Driving both versions side by side over 2100 random key sequences (digits, backspace, clearLine) across seven patterns, the buffer contents, `patternArrayPos`, cursor column and rendered field agreed on every one. Thirty-five lines become fourteen, and the `assert` and the unbounded `while (this.patternArrayPos >= 0)` go with them.

### One backspace arm is untested

`backspace after clearText() stays aligned with the text` steps from four filled slots back across the `/`, which is the literal walk-back. A second backspace takes the other arm, where the cursor is already on an input slot. Suggestion inline.

### Pre-existing, and adjacent enough to note

A view constructed with `text:` in its options never loads the buffer. `TextView`'s constructor calls `setText` at `core/text_view.js:52`, before `buildPattern` runs and before `lineBuffer` exists, so both guards in the override skip:

```
new MaskEditTextView({ maskPattern: '####/##/##', text: '1990' })
  this.text        = "1990"      <- drawn
  lineBuffer chars = ""
  getData()        = ""
```

The field shows `1990` and returns nothing, and the first keypress writes over it. master behaves identically, so this branch is not a regression. It is worth knowing here because the new `this.patternArrayPos = this._patternPosForLength(0)` in the constructor is right only while the buffer starts empty.

test/views.test.js
andy5995: This should be in 672269cd

andy5995:
I opened this because when I was applying, the cursor became out of focus (was over to the right about 6 columns) when entering birthdate after I had started entering in "20" and then hit backspace. I reproduced it a second time when setting up the board locally and adding myself as a new user. But I tried reproducing it on Enigma through ssh at the command line and when logged in with syncterm, but I could not.

NuSkooler:
This is great, thank you!

PR #783, opened 2026-09-06
fix(menus): three defects in the shipped menu templates

NuSkooler:
Two keys in the shipped menu templates that have no effect. Found while establishing the legal key set for `menu.hjson` validation — neither appears in `docs/_docs/configuration/menu-hjson.md`, and neither is read anywhere in `core/`.

## `userSSHKeyConfig` had a menu level `acs`

```hjson
userSSHKeyConfig: {
    desc: SSH Public Keys
    module: user_ssh_key_config
    art: SSHKEYCFG
    acs: SC          //  never read
    form: { ... }
}
```

The only gate on *entering* a menu is `hasMenuModuleAccess()` (`core/acs.js:83-94`), called from `core/menu_stack.js:138`:

```js
const acs = _.get(modInst, 'menuConfig.config.acs');
if (!_.isString(acs)) {
    return true;    //  no ACS check req.
}
```

It reads `menuConfig.config.acs`. A menu level `acs` is consulted by nothing. Moved into `config{}`.

### Scope, stated honestly

**A stock install was not exposed.** The route from the main menu is separately gated by an *action* level `acs: SC` on the SSH command (`main.in.hjson:194`), and action ACS does work:

```hjson
{
    //  Allow SSH key management only from secure transports
    acs: SC
    action: @menu:userSSHKeyConfig
}
```

What was missing is the defence-in-depth line behind it. Any other route to the menu — a sysop's custom menu, an `@menu:userSSHKeyConfig` from elsewhere — reached it with no transport check, on a template whose evident intent is "secure transports only". Low severity; worth having actually work.

## `newUserApplicationSsh` had `fallback: logoff`

Nothing has read a menu `fallback` since `f7a7423b` replaced the old fallback machinery with the menu stack (`@systemMethod:fallbackMenu` → `@systemMethod:prevMenu`). The comment above it —

```
//  Canceling this form logs off vs falling back to matrix
```

— described behaviour the key no longer produces, so it goes too.

## Why now

Both are prerequisites for the `menu.hjson` validation work being built next, not optional tidying. The acceptance criterion inherited from #281 is that a configuration produced by `oputil.js config new` yields **zero** unknown-key findings; a validator closing the menu entry key set on what the code actually reads would flag both of these on a brand new install.

Kept separate so a security-adjacent fix is not buried inside a large validator PR.

## Testing

`npm test` — 2316 passing, live suite included. All templates parse; `oputil.js config new` output unaffected beyond the two keys.

### DIFF
--- a/misc/menu_templates/activitypub.in.hjson
+++ b/misc/menu_templates/activitypub.in.hjson
@@ -14,7 +14,7 @@
           action: [
             {
               acs: "AE1"
-              action: @menu:activityPubActorSearch
+              action: @menu:activityPubSearch
             }
             {
               action: @menu:activityPubNotEnabled
--- a/misc/menu_templates/main.in.hjson
+++ b/misc/menu_templates/main.in.hjson
@@ -605,7 +605,11 @@
             desc: SSH Public Keys
             module: user_ssh_key_config
             art: SSHKEYCFG
-            acs: SC
+            config: {
+                //  Only core/acs.js hasMenuModuleAccess() gates a menu, and it
+                //  reads config.acs -- a menu level "acs" is never consulted.
+                acs: SC
+            }
             form: {
                 0: {
                     mci: {
--- a/misc/menu_templates/new_user.in.hjson
+++ b/misc/menu_templates/new_user.in.hjson
@@ -125,13 +125,10 @@
 
         //
         //  SSH specialization of NUA
-        //  Canceling this form logs off vs falling back to matrix
-        //
         newUserApplicationSsh: {
             desc: Applying
             module: nua
             art: NUA
-            fallback: logoff
             next: newUserFeedbackToSysOpPreamble
             form: {
                 0: {


NuSkooler:
## Added: the ActivityPub search command is broken

Found by pointing the `menu.hjson` reference checking (PR A of the follow-on work) at the shipped templates for the first time.

`misc/menu_templates/activitypub.in.hjson:17` routes the `S` command to a menu that does not exist:

```hjson
{
  value: {command: "S"}
  action: [
    {
      acs: "AE1"
      action: @menu:activityPubActorSearch     //  no such menu, anywhere
    }
    ...
```

Meanwhile `activityPubSearch` — module `activitypub/ap_search` — **is** defined at `:111` and referenced by nothing.

So on a stock install with ActivityPub enabled, pressing `S` in the ActivityPub menu fails with `No menu entry for "activityPubActorSearch"` instead of opening search, and the search menu that does exist is unreachable. One word.

After the fix there are **no dangling `@menu:` references anywhere in the templates**.

This is the third template defect in this PR and the first one a user would actually hit, so I have retitled it accordingly. 2316 tests still pass.

Issue #779, opened 2026-09-06
Pre-auth feedback shipped without its menu template


NuSkooler:
Pre-auth feedback (#695, `2a5c60d5`, 2026-05-20) shipped its module, its documentation and its theme block, but **not its menu template**. A fresh install therefore cannot enable the feature by following the documentation.

## What is missing

`git log -S preAuthFeedback -- misc/menu_templates/` returns nothing. No file under `misc/menu_templates/` mentions `preAuthFeedback` at all:

```
$ grep -rln "preAuthFeedback" misc/menu_templates/
(none)
```

Everything else from that PR is present:

| | |
|---|---|
| `core/pre_auth_feedback.js` | ✅ |
| `docs/_docs/modding/pre-auth-feedback.md` | ✅ |
| `art/themes/luciano_blocktronics/theme.hjson` → `customization.menus.preAuthFeedback` | ✅ |
| `misc/menu_templates/login.in.hjson` → `preAuthFeedback` menu | ❌ |
| `misc/menu_templates/login.in.hjson` → the `VM1` matrix item that reaches it | ❌ |

## Why it matters

The documentation points at a template that does not contain the example it names:

> `docs/_docs/modding/pre-auth-feedback.md:18`
> 2. Add a `preAuthFeedback` menu entry to your login menu config (see `misc/menu_templates/login.in.hjson` for a complete example).

A sysop following those instructions on a config generated by `oputil.js config new` finds no example to copy. Step 1 of the same list — wiring a `feedback` item into the login matrix's `VM1` — is likewise absent.

The knock-on effect is that the shipped `luciano_blocktronics` theme carries a `preAuthFeedback` customization that is dead on any stock install, since `ThemeManager` only consults theme entries for menus that exist. That block is correct and is being **left in place** by #778 precisely because this is a template gap rather than a stale theme entry.

## Suggested fix

Add to `misc/menu_templates/login.in.hjson`:

1. The `preAuthFeedback` menu entry, matching the MCI layout the theme block already expects — form 0 `ET1`/`ET2`/`ET3` and form 1 `MT1`, the same shape as `privateMailMenuCreateMessage`, which the docs already suggest copying from.
2. The `feedback` item in the login matrix `VM1` items list, wired to `@menu:preAuthFeedback`.

Worth checking the other menus that PR added, if any, at the same time.

## Broader point

[...]

PR #784, opened 2026-09-06
feat(config): validate menu.hjson and theme.hjson

NuSkooler:
PR A of the successor to #281. Design and sizing in the vault; follows #778 (theme fix) and #783 (template fixes), both prerequisites.

`menu.hjson` and `theme.hjson` carry the same silent failure `config.hjson` did, and one edge is worse than anything there. `ThemeManager._finalizeTheme()` starts from `menu.hjson` and iterates **the menu's** keys, looking each up in the theme:

```js
const mergedTheme = structuredClone(this.menuConfig.get());     //  core/theme.js:193
Object.keys(mergedTheme[sectionName]).forEach(entryName => {
    const menuTheme = _.get(theme, ['customization', sectionName, entryName]);
```

So a customization naming a menu that does not exist is **never consulted**. The theming simply does not happen, with nothing logged. The theme we ship had eight such blocks (#778).

## Structure

`core/config/menu_schema.js` and `core/config/theme_schema.js`, hand written — `core/theme.js` hands its `ConfigLoader` no `defaultConfig`, so there is nothing to derive from.

Only the frame is described. **`config{}`, `form{}` and MCI blocks stay wide open**: `config{}` carries 69 module-specific keys across 56 modules on a real board, and `form`/`mci` needs a grammar that is its own piece of work. Claiming to know either would report most menus on most boards.

### The key sets come from the code, and both documents would have misled me

| | |
|---|---|
| `font`, `runtime`, `youSubmittedFormat` | legal, read at `menu_module.js:906`, `menu_module.js:644`, `bbs_list.js:213` — listed in no menu documentation |
| `statusAvailableIndicators` | the real key; `docs/_docs/art/themes.md` lists `getStatusAvailIndicators`, which is the **helper method** name (`core/theme.js:386`) |
| `info.desc` | read at `core/user_config.js:218`, undocumented |

Closing on either document alone would have reported working boards as broken.

## References

`validateMenuReferences()` checks `next`, `prompt`, and **every `@menu:` anywhere beneath a menu entry** — found by walking rather than by modelling `submit` and `actionKeys` shapes, which vary enough (a string, an array of ACS alternatives, an array of blocks each with their own action) that enumerating them would be its own source of misses. That is how the broken ActivityPub search command in #783 was found, three levels down inside an ACS-guarded action array.

A leading `@` makes a value an *asset spec*, not a menu name (`core/asset.js:20-34`), so only bare values and explicit `@menu:` are checked. A first cut reported four dangling menus on a production board — all `@systemMethod:logoff`, all correct.

`validateThemeReferences()` reports a customization naming no menu or prompt under a new `deadCustomization` code at **warning** severity. A board running a menu file older than its theme gets these legitimately; failing its `ExecStartPre` over cosmetics would be wrong.

## Surfaces

- **`menu.hjson`** through its own `ConfigLoader` hook.
- **Themes** in a single pass after `_loadThemes`, not per theme: the interesting check is cross-file and a per-theme hook cannot see `menu.hjson`. §5 of the #281 design predicted this hook point.
- **`oputil.js config validate`** covers both, naming each of the four files it can now report on.

Both guarded — a throw would otherwise escape into the startup series whose callback would then never fire. Both to the log, since menus and themes load well after `Log.init()`.

`misc/menu.schema.json` is published alongside the configuration artifact from the same generic emitter, with the same staleness and reproducibility guards.

## Measured, not assumed

| | Result |
|---|---|
| Stock templates (119 menus, 14 prompts) | **clean** — structure and references |
| Dev install (134 menus) | 2, both dead keys #783 removed |
| **Production board (253 menus)** | **1** — an inherited dead key. **No false positives.** |
| That board's own maintained theme | **zero** orphans |

Live boot with a deliberately broken menu file and theme reports all five findings, each labelled by source file, board starts regardless.

```
40 | dev_enig-main.hjson: unknown key "acs"
40 | dev_enig-main.hjson: unknown key "fallback"
50 | dev_enig-main.hjson: menu "prevMenu" is not defined in menus
40 | broken_theme/theme.hjson: unknown key "getStatusAvailIndicators"
40 | broken_theme/theme.hjson: no menu named "noSuchMenuAtAll" -- this customization is never applied
```

## Testing

`npm test` — **2352 passing** (up from 2316), live suite included. `npm run lint` and `prettier --check` clean. Both artifacts reproduce byte-identically.

## Not in scope

`config{}` per-module key sets, `art` existence, and MCI/`form{}` validation — all logged with complexity and gain in the vault design. The last is the largest remaining win and needs its own effort.


NuSkooler:
## Independent review — two real findings, both fixed in `461f2853`

### 1. `action` is a legal menu entry key, and the schema rejected it

The important one: a **false positive on a working menu**, which is the failure mode this whole effort is supposed to avoid.

A menu driven by a `prompt` may carry its action directly rather than in a by-form-id submit block. `core/view_controller.js:653` reads it straight off the entry, and the `else` branch immediately below spells out why:

```js
if (_.isString(this.client.currentMenuModule.menuConfig.action)) {
    handleIt(formData, this.client.currentMenuModule.menuConfig);
} else {
    //  Menus that reference prompts can have a special "submit" block without the
    //  hassle of by-form-id configurations, etc.
```

Confirmed by construction through the real CLI:

```hjson
confirmLogoff: {
    desc: "Confirm Logoff"
    prompt: confirmLogoffPrompt
    action: @systemMethod:logoff
}
```
```
menu.hjson: 1 issue (1 warning)
  warning  menus.confirmLogoff.action
           unknown key "action"
```

So it is a **fourth** undocumented-but-legal key, not the three the module header claimed. My own audit grep saw `menuConfig.action` once and I wrote it off as belonging to `actionKeys`. Now accepted, with a test.

### 2. Stale help text

`oputil.js config validate` still described itself as covering `config.hjson` and `achievements.hjson`, which this PR made untrue.

## Verified clean by the review

Constructed and run, not merely read: the `theme.hjson` key sets against `core/theme.js` and `core/user_config.js` (including the docs listing helper names where the config keys differ); the `@` asset-spec handling against `core/asset.js` for every type plus `@menu:location:name`, names with `.`/`/`, and malformed forms; all three `next` shapes; the `@menu:` walk for false-positive sources; ~35 malformed inputs fuzzed through the validators with zero throws; callback and `process.exitCode` discipline in the restructured oputil nesting; and an instrumented `Module._load` confirming nothing on the `config validate` path pulls in `stat_log.js` or `database.js`.

## Known, and tracked elsewhere

A fresh `oputil.js config new` install with the default theme reports one warning on day one — `customization.menus.preAuthFeedback`. That is #779: the art and the theme block ship, the menu template does not. It is a true positive pointing at a real gap rather than a validator bug.

2353 passing, lint and prettier clean, both artifacts reproduce.

PR #785, opened 2026-09-06
feat(config): check that a menu's module can actually be loaded

NuSkooler:
PR B of the menu.hjson work, following #784. Completes the menu/theme validation sized in the vault.

A menu's `module` names something that has to be loadable. `core/config/module_resolver.js` mirrors what `core/menu_util.js:79-90` asks of the value and what `core/module_util.js` `loadModuleEx()` then does with it:

- a bare name is a system module under `core/`, `@userModule:` names one under `paths.mods`;
- either may live at `<name>.js` or, so a module can carry its own `package.json`, at `<name>/<name>.js`.

**A resolver rather than a name set**, because a module name may be a relative path — `activitypub/ap_search` and `./activitypub/activity_pub_msg_list` are both in the shipped templates — so enumerating names would mean reimplementing the lookup instead of asking it. It also refuses a name that resolves outside its own directory, since separators are legal in one.

## It found a broken sysop command on its first run

`sysopBinkpPollNow` names `module: binkp_poll`. **No such module has ever existed.** The real one is `core/binkp/binkp_poll_module.js` — *BinkP Poll, "Trigger an immediate BinkP outbound poll"*.

And the menu is reachable. `main.in.hjson:164-171`:

```hjson
{
    value: { command: "!BINKP" }
    action: [
        {
            acs: SU
            action: @menu:sysopBinkpPollNow
        }
    ]
}
```

So the sysop `!BINKP` command has failed with `MODULE_NOT_FOUND` since #682 added it. One word, fixed here — the templates have to be clean for the check to be shippable.

## Two things are reported

A module that is not there, and a value that is **not a module spec at all**. The second matters more than it looks: `getModuleAsset()` asserts the type is `systemModule` or `userModule`, and an assert throws straight out of the waterfall that loads the menu — so `@method:foo` in a module position is worse than a missing file.

```
error    menus.missing.module
         menu module "no_such_module_at_all" is not defined in the core modules
         -- no "no_such_module_at_all.js" there, nor "no_such_module_at_all/no_such_module_at_all.js"

error    menus.wrongType.module
         menu module "@method:notAModule" is not defined in the core modules
         -- "@method:" does not name a module; use a bare name or "@userModule:"
```

The spec is parsed in `refs.js` rather than by requiring `core/asset.js`, which pulls in `stat_log.js` and its load-time database handle — that has no business happening because somebody validated a configuration. A test asserts the copy still agrees with `core/asset.js` on every form.

## A warning flood, prevented

Pointing `oputil config validate` at a minimal menu file produced **66 dead-customization warnings** against the stock theme. Every line was true, and the result was unusable — the exact "wall of text, so it gets switched off" outcome #281's plan warns about.

A theme whose customizations *mostly* name nothing is now reported once:

```
warning  customization.menus
         66 of 66 menu customizations name something this system does not define,
         so they are never applied -- this theme looks written for a different menu file
```

A handful of dead entries are typos worth naming individually; seventy is a different fact.

## Measured

| | Result |
|---|---|
| Stock templates | clean after the `binkp_poll` fix |
| **Production board, 253 menus** | **all 56 modules resolve** — including its own `mods/` and eight `@`-form specs. Nothing reported. |

## Testing

`npm test` — **2372 passing**, live suite included. `npm run lint` and `prettier --check` clean; both artifacts reproduce.

Includes a crash fix found by the review of #784: `module: ""` made the parser return `undefined` and the next line read `.type` off it, escaping oputil's nested callbacks and killing the process with a raw stack trace. The parser now never returns `undefined`, and `''`, `'   '`, `'@'`, `'@@@'`, `'@:'`, `'@menu'` are each reported rather than thrown.

### DIFF
--- a/WHATSNEW.md
+++ b/WHATSNEW.md
@@ -40,6 +40,8 @@ This document attempts to track **major** changes and additions in ENiGMA½. For
 
   **`menu.hjson` and your themes are checked too.** They carry the same silent failure as `config.hjson`, and one of them is worse: `ThemeManager` starts from your menu file and looks each menu up in the theme, so a `customization` naming a menu that does not exist is **never consulted** — the theming simply does not happen, with nothing logged. The theme ENiGMA½ ships had eight such blocks; they are fixed, and a check now reports any that remain as a warning.
 
+  **A menu naming a module that cannot be loaded is reported too**, for both built-in modules and your own under `paths.mods`. That found the sysop `!BINKP` command pointing at `binkp_poll`, which has never existed — the module is `binkp/binkp_poll_module` — so the command has failed since native BinkP support landed.
+
   Menu files get their structure checked — a misspelled `modul` for `module`, an `acs` where only `config.acs` is read — along with every menu name they reference, wherever it appears: `next`, a submit handler, an action key. That last one immediately found the `S` command in the ActivityPub menu pointing at a menu that has never existed.
 
   `config{}`, `form{}` and MCI blocks are deliberately **not** checked: those belong to the modules that read them and to the art they position, and claiming to know them would report most menus on most boards.
--- a/core/config/issue.js
+++ b/core/config/issue.js
@@ -139,6 +139,13 @@ function describeIssue(issue) {
             break;
 
         case IssueCodes.DeadCustomization:
+            if (undefined !== issue.count) {
+                message =
+                    `${issue.count} of ${issue.total} ${issue.refKind} customizations name something this system does not define, ` +
+                    'so they are never applied -- this theme looks written for a different menu file';
+                break;
+            }
+
             message = `no ${issue.refKind} named "${issue.value}" -- this customization is never applied`;
             if (issue.suggestion) {
                 message += `; did you mean "${issue.suggestion}"?`;
--- a/core/config/module_resolver.js
+++ b/core/config/module_resolver.js
@@ -1,0 +1,78 @@
+/* jslint node: true */
+'use strict';
+
+//  deps
+const fs = require('graceful-fs');
+const paths = require('path');
+
+//
+//  Can a menu's "module" actually be loaded?
+//
+//  Mirrors what core/menu_util.js:79-90 asks of it and what
+//  core/module_util.js loadModuleEx() then does with the answer:
+//
+//    * the value goes through asset.getAssetWithShorthand(spec,
+//      'systemModule'), so a bare name is a system module and "@userModule:"
+//      names one under Config().paths.mods;
+//
+//    * a module may live at <path>/<name>.js or, so that it can carry its own
+//      package.json and dependencies, at <path>/<name>/<name>.js.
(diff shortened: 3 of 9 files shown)


NuSkooler:
## Independent review — three findings, all fixed in `29b08d81`

### 1. `oputil.js config validate` had no guard at all (crash)

The command called the validators directly, with no try/catch anywhere in the chain. The boot and hot-reload paths are covered by `ConfigLoader._validate()`'s own guard; this one was not — so a single throw killed the process with a raw stack trace, **skipped every remaining file**, and left `process.exitCode` as Node's uncaught-exception code rather than the one the command meant to set.

That gap predates this PR — it arrived with the command in #766 — but this PR is what made it reachable, so it is fixed here. Each file's validation now runs through `collectIssues()`, which names the file it could not check and carries on.

Verified by injecting a throw into the resolver and running the real command. Before: a stack trace, nothing else checked. After:

```
  (could not check .../menu.hjson: injected)

.../theme.hjson: 1 issue (1 warning)
  ...
.../achievements.hjson: no problems found
```

### 2. The containment check only looked at one of the two candidates (missed detection)

`module_resolver.js` checked `<name>.js` for containment but not `<name>/<basename>.js`. Given `..`, appending `.js` makes a filename *inside* the root — so that candidate looks contained — while the nested form climbs out of it. The escape happened on the candidate nobody was looking at:

```
".."          direct in root: true   | contained in root: false
"./.."        direct in root: true   | contained in root: false
"foo/../.."   direct in root: true   | contained in root: false
"../package"  direct in root: false  | contained in root: false
```

Worse, the test named *"refuses a name that climbs out of its directory"* never covered this, because `../package` escapes on the **first** candidate — so it did not prove what its name claimed. Both candidates are now checked, and the test plants a file above `core/` so it actually exercises the case.

### 3. A boxed `String` threw (crash, unreachable through hjson)

`_.isString()` — the gate in `refs.js` — is true for a boxed `String`, which `path.basename()` then refuses outright. Not reachable through `hjson.parse()`, which yields primitives, but it is the concrete mechanism that made finding 1 a crash rather than a report. The name is now coerced.

## Verified clean by the review

Constructed and run: ~35 spec forms against the real `getAssetWithShorthand` (one deliberate, harmless classification difference on the empty string); the aggregation threshold boundaries with `count`/`total` correctness at 4/2, 4/3, 8/4, 8/5, 100/51, 1000/501; `core/theme.js`'s wiring proven safe through `ConfigLoader`'s existing try/catch with a deliberately throwing validator; an instrumented `Module._load` confirming `stat_log.js` and `core/asset.js` are never loaded on this path and no database is opened; and the full shipped template set end-to-end reporting "no problems found".

The reviewer also confirmed the `enabled: false` divergence is benign — the resolver does not model runtime enable state, and disabling a mod is a deliberate choice rather than a broken reference.

2374 passing, lint and prettier clean.

PR #787, opened 2026-09-07
fix(doors): write the drop file comm fields the door is actually given

NuSkooler:
Supersedes #781, whose commit is included here unchanged and with authorship intact — thanks @andy5995. The original fix is right; this adds the three things it left on the table.

## The bug (#781)

`getDoor32Buffer` hardcoded comm type `2` (telnet) with handle `-1`, carrying a `:TODO:` since 2016. That pair is not a state the [spec](https://raw.githubusercontent.com/NuSkooler/ansi-bbs/master/docs/dropfile_formats/door32_sys.txt) permits — its sample local caller is `0` / `0` — and a door that checks it is entitled to refuse to start. [Immortal Barons](https://github.com/andy5995/immortal-barons) does exactly that:

```go
// Socket mode with no handle would attach to fd 0 (stdin) instead of the
// caller — fail rather than talk to the wrong stream.
if c.IO == IOSocket && sock <= 0 {
    return nil, fmt.Errorf("line 2 (socket handle): socket mode needs a handle, got %d", sock)
}
```

## What this adds

**`commType` is a config key, not just an `io` derivation.** `io` says how ENiGMA½ talks to the process it spawns; the drop file says how the *door* talks to the caller. Those coincide only when the process we spawn **is** the door. Put QEMU in between — bridging `{srvPort}` onto an `isa-serial` line — and the guest sees a COM port, so `DOOR32.SYS` wants serial and would still have been handed telnet with a `-1` handle: the very pair #781 set out to stop writing. The `io`-derived value stays as the default, so nothing changes for a config that does not set the key; an unrecognized value warns and falls back rather than quietly reporting something else.

**`DOOR.SYS` and `DORINFO` had the same defect.** Line 1 was hardcoded `COM1:` under a comment reading `"Comm Port - COM0: = LOCAL MODE"`, and DORINFO line 4 hardcoded `COM1` under `"or 0 if logged in on console"`. Both now follow `commType`.

**`DropFile` normalizes `commType` on the way in**, so the DOOR32 map no longer needs a fallback arm for a value that cannot reach it.

**One test asserted the opposite of its name.** `never claims a socket while supplying no handle` passed on `-1` — which *is* no handle. ENiGMA½ cannot write an honest telnet pair at all: it shares a socket *server*, not a descriptor, and [bivrost!](https://github.com/NuSkooler/bivrost) overwrites both lines with the real fd (`write_new_door32sys_dropfile` writes `2\r\n` then the descriptor, copying only lines 3+). The test now pins that contract instead of pretending the `-1` is usable.

Docs and the `doors.in.hjson` template cover the new key.

## What changes for existing boards

| Setup | Before | After |
|---|---|---|
| `io: stdio` (the default) + `DOOR32` | `2` / `-1` — invalid | `0` / `0` |
| `io: stdio` + `DOOR.SYS` / `DORINFO` | `COM1:` / `COM1` | `COM0:` / `0` |
| `io: socket` + bivrost! | `2` / `-1` | unchanged, and overwritten by bivrost! either way |
| `io: socket`, no bivrost! | `COM1:` / `COM1` | unchanged; set `commType: serial` for an emulator bridge |
| `v86_door` | `2` / `-1`, `COM1:`, `COM1` | `1` / `0`, `COM1:`, `COM1` |

The DOOR32 change is risk-free — the old value was invalid, so nothing working could depend on it. The DOOR.SYS/DORINFO change only touches local doors, and both shipped templates are non-local and keep `COM1`: `doorAbracadabraExample` is `io: socket`, and `doorMyGameV86` loads a FOSSIL driver and writes to `COM1`, which is precisely what `serial` reports.

## Verification

Full suite green (2387 passing), prettier and eslint clean. 13 tests in `test/dropfile_comm_type.test.js` cover all three formats across `local` / `serial` / `socket` / unset / unrecognized / mixed case.


test/dropfile_comm_type.test.js
```diff
@@ -0,0 +1,108 @@
...
+
+    //
+    //  ENiGMA½ has no descriptor to share, so it cannot write an honest
+    //  telnet/handle pair at all: bivrost! overwrites both lines with the
+    //  real fd before the door reads the file. This pins the contract rather
+    //  than pretending the -1 is something a door could use.
+    //
+    it('leaves the socket pair for bivrost! to rewrite', () => {
+        assert.deepEqual(door32Lines('socket').slice(0, 2), ['2', '-1']);
+    });
```
andy5995, line 66: `leaves the socket pair for bivrost! to rewrite` asserts `['2', '-1']`, which `keeps the telnet spelling when a socket is shared` already asserts above it. The comment is the part that carries information, so it can move onto the first test:

```suggestion
    //
    //  ENiGMA½ has no descriptor to share, so it cannot write an honest
    //  telnet/handle pair at all: bivrost! overwrites both lines with the
    //  real fd before the door reads the file. This pins the contract rather
    //  than pretending the -1 is something a door could use.
    //
    it('keeps the telnet spelling when a socket is shared', () => {
        const lines = door32Lines('socket');
        assert.equal(lines[0], '2', 'comm type should be telnet');
        assert.equal(lines[1], '-1');
    });

    it('reports serial for a door on an emulated COM port', () => {
        const lines = door32Lines('serial');
        assert.equal(lines[0], '1', 'comm type should be serial');
        assert.equal(lines[1], '0', 'a serial port carries no socket descriptor');
    });
```

The alternative is to keep both and have this one assert what makes it different — that lines 3+ are what bivrost preserves.

andy5995 (commented):
> Written by Claude (Opus 5), an LLM made by Anthropic, at andy5995's direction.

I ran `DropFile` directly against the tree at this branch for all three formats. The output matches the table in the description:

```
DOOR32   local / unset / unrecognized -> 0,0      serial -> 1,0            socket -> 2,-1
DOOR     local / unset / unrecognized -> COM0:    serial / socket -> COM1:
DORINFO  local / unset / unrecognized -> 0        serial / socket -> COM1
```

Three findings.

### The DOSEMU example changes behaviour and nothing covers it

`docs/_docs/modding/local-doors-dos-emulation.md` sets `$_com1 = "virtual"`, loads BNU in `AUTOEXEC.BAT`, and its menu entry is `io: stdio` with `dropFileType: DORINFO`. The guest there sees COM1, which is the v86 situation — but `io: stdio` now defaults to `local`, so DORINFO line 4 goes from `COM1` to `0` on that documented setup. It needs `commType: serial` for the same reason v86 does, and that page is where a sysop would look.

The QEMU entry below it is the case the description names — `-device isa-serial,chardev=s0` bridging `{srvPort}` — and does not set the key either. That one is `io: socket`, so DOOR.SYS line 1 stays `COM1:` and nothing breaks; it is just the one example a reader would expect to see `commType: serial` in.

### `io: socket` still defaults to the `2`/`-1` pair

bivrost!'s `write_new_door32sys_dropfile` writes its own `2\r\n` and fd and copies only from line 3, so with bivrost in place `commType: socket` never reaches the door for DOOR32. Where it does reach a door is the case bivrost is absent — a native door dialling back to `{srvPort}` — and there it writes the pair the first commit set out to remove. `normalizeCommType`'s own comment states the argument against it: local is "the one mode that asks nothing of us, and so cannot promise a door something we are unable to hand over."

Defaulting `io: socket` to `local` would cost a bivrost board nothing on DOOR32, and would let Immortal Barons (or anything else validating the pair) start under a plain socket door. It costs DOOR.SYS and DORINFO, which bivrost does not rewrite: `COM1:` becomes `COM0:` for a bivrost door that reads those, and that door does hold a descriptor, so local is the wrong thing to tell it. I lean toward the local default with `commType: socket` left available for boards that want the old bytes, but the DOOR.SYS side is a real argument the other way and the call is yours.

### No WHATSNEW.md or UPGRADE.md entry

`commType` is a new config key, and the DOOR.SYS and DORINFO defaults move for every existing `io: stdio` door. Both files carry that level of detail for 0.5.1-beta.

### Minor

- `core/abracadabra.js:91` repeats the lowercase-and-check that `DropFile.normalizeCommType` performs. Returning null from it for an unrecognized value would let the warn path reuse it.
- `core/dropfile.js:222` — "'serial' reports 0, which door libraries take as the first port" is not in `door32_sys.txt`, which gives no value for line 2 under comm type 1. It is the common convention; the spec is yours to extend if you want it pinned there.

NuSkooler:
Thank you for the bug report and inspiration @andy5995! Hope you don't mind that I stole some of your code here :D

andy5995:
> Thank you for the bug report and inspiration @andy5995! Hope you don't mind that I stole some of your code here :D

Nah, you added my commit and maintained the co-authored trailer. Not theft in my book.

PR #788, opened 2026-09-07
ci: cancel superseded pull request runs

andy5995:
> direction.

Test, Lint and CodeQL run on pull_request and set no concurrency, so a fixup
pushed to a branch leaves the superseded runs to finish. Grouping on the pull
request number where there is one and the commit sha otherwise gives every push
its own group, so a master push is never cancelled by the next; only pull
request runs supersede each other. Docker and Jekyll have no pull_request
trigger and so nothing to supersede, and are left alone.

https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/control-workflow-concurrency

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -20,6 +20,10 @@ on:
   schedule:
     - cron: '41 0 * * 5'
 
+concurrency:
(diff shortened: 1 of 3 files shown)


NuSkooler (approved):
ty!

PR #796, opened 2026-09-09
dropfile: add BBSDEV.DRP, discovered through the BBSDEV_DRP env var

andy5995:
This is a new dropfile format created by the maintainer of syncterm, @RealDeuce (see https://realdeuce.github.io/bbsdev.drp/)

Some new doors have support for this already, such as [Immortal Barons](https://andy5995.github.io/immortal-barons/) (my project) and [Family Feud](https://bbswordle.com/feud/).


NuSkooler:
Paging @RealDeuce !

core/dropfile.js
```diff
@@ -279,7 +421,83 @@ module.exports = class DropFile {
...
+    //  See https://github.com/RealDeuce/bbsdev.drp for the specification,
+    //  its ABNF grammar, and an example of each communications mode.
+    //
+    //  The door finds this file through the BBSDEV_DRP environment variable
+    //  rather than an argument; abracadabra sets it.
+    //
+    getBbsDevBuffer() {
+        const user = this.client.user;
+        const term = this.client.term;
+        const encoding = term.outputEncoding || 'cp437';
```
NuSkooler, line 435: **Line 12 names the terminal's encoding, but ENiGMA decodes the door with `config.encoding`**

ENiGMA transcodes door output rather than passing it through. `door.js` decodes whatever the door writes using `exeInfo.encoding` — the door's `config.encoding`, defaulting to `cp437` — and only then re-encodes for the terminal:

```js
// core/door.js:58
this.encoding = (exeInfo.encoding || 'cp437').toLowerCase();
// core/door.js:234
doorDataHandler(data) { this.client.term.write(decode(data, this.encoding)); }
```

Line 12 here is built from `term.outputEncoding` instead, so the two disagree whenever the sysop's `encoding` isn't the caller's terminal encoding. With a Linux/xterm caller against a stock door:

```
drop file line 12 (encoding told to the door): "UTF-8"
door.js decodes the door's output with exeInfo.encoding = "cp437"
```

The door obeys line 12, emits UTF-8, ENiGMA decodes it as CP437, and the caller gets mojibake. It inverts for a SyncTERM caller against an `encoding: utf8` door. This is the field WHATSNEW puts forward as the reason to adopt the format, so it's worth getting exactly right.

The fix looks local to the two files already touched here: `abracadabra` reads `config.encoding` at line 263 to build `exeInfo`, and could pass the same value into the `DropFile` options at line 228 alongside `commType`/`commParams`. Line 12 then states what ENiGMA will actually do with the door's bytes.

Two things to watch if you go that way. `bbsDevError()` at line 489 checks `term.outputEncoding` too, and would need to check whichever value the buffer ends up writing so the guard and the field can't diverge. And since an encoding missing from `BbsDevEncodingNames` is a hard refusal to launch, sourcing line 12 from `config.encoding` changes which values have to be in that table — a door configured `encoding: 437` (a valid iconv alias for cp437) would stop starting.

core/dropfile.js
```diff
@@ -279,7 +421,83 @@ module.exports = class DropFile {
...
+
+        return Buffer.from(
+            [
+                '1.0', //  format version
+                this.commType,
+                this.commParams, //  empty for 'local' and 'stdio'
+                bbsDevField(user.username, `user${user.userId}`),
+                user.userId.toString(), //  opaque, stable, ours alone
```
NuSkooler, line 443: **The user key satisfies "stable" but not "never reassigned"**

The spec asks two things of line 5: a producer "MUST keep it stable for the lifetime of the user account and MUST NOT assign it to another account on the same BBS." The comment on this line — `opaque, stable, ours alone` — holds for the first, but `userId` can't promise the second.

`user.id` is declared `INTEGER PRIMARY KEY` without `AUTOINCREMENT`, which makes it a plain rowid alias, and SQLite reuses the largest unused rowid. Remove the highest-id user with `oputil user remove` and the next account created inherits that id. A door keying persistent state on the user key — scores, saves, per-user settings — would then serve the deleted user's data to a brand-new user, silently.

Flagging this so the choice is deliberate rather than assumed, not to ask for a schema change: the user table is far outside this PR, and changing how ENiGMA allocates user ids isn't on the table. Whatever satisfies the spec here needs to live in `getBbsDevBuffer()` and derive a key from what's already on the user.

One caveat if you reach for the obvious candidate. Composing `userId` with `UserProps.AccountCreated` would distinguish a recycled id, but that property is written in exactly one place — `core/nua.js:108`, the new-user application flow — so accounts created by `oputil user add` don't have it, which likely includes the first sysop. `user_login.js:257` already guards `if (created)` for that reason. Anything built on it needs a defined answer for users where it's absent.

core/dropfile.js
```diff
@@ -279,7 +421,83 @@ module.exports = class DropFile {
...
+                (term.termWidth || 80).toString(),
+                (term.termHeight || 25).toString(),
+                'Y', //  ANSI: every ENiGMA½ session is drawn with it
+                'N', //  RIP: not supported
+                term.ctermVersion || '', //  empty unless the caller answered DA as CTerm
+                '', //  time of logoff: ENiGMA½ has no per-call time limit
+                bbsDevEncodingName(encoding),
+                bbsDevField(Config().general.language, 'en-US'),
```
NuSkooler, line 451: **`general.language` isn't validated as a BCP 47 tag**

Line 13 MUST be a well-formed BCP 47 language tag, but `bbsDevField` only strips control characters and trims — a sysop who sets `language: "English (US)"` gets it written verbatim, and a conforming door must then reject the whole file.

This is also inconsistent with how the PR treats line 12 three lines up: an encoding that can't be named in the registry's spelling is a hard refusal with a logged reason, which is the right call. A malformed language tag produces a file that's just as invalid, silently.

Either validate well-formedness and fold it into `bbsDevError()` alongside the encoding check, or fall back to `en-US` when it doesn't parse. The `general.language` entry this PR adds to `misc/config.schema.json` is another spot that could carry a `pattern`, which would at least surface it through `oputil config validate` at startup rather than at door launch.

core/dropfile.js
```diff
@@ -279,7 +421,83 @@ module.exports = class DropFile {
...
+        return Buffer.from(
+            [
+                '1.0', //  format version
+                this.commType,
+                this.commParams, //  empty for 'local' and 'stdio'
+                bbsDevField(user.username, `user${user.userId}`),
+                user.userId.toString(), //  opaque, stable, ours alone
+                (term.termWidth || 80).toString(),
+                (term.termHeight || 25).toString(),
+                'Y', //  ANSI: every ENiGMA½ session is drawn with it
```
NuSkooler, line 446: **ANSI is hardcoded `Y` though the terminal may say otherwise**

`ClientTerminal.prototype.isANSI()` already exists at `client_term.js:157` and checks the terminal type against `term.cp437TermList`. It'd be false for several types ENiGMA accepts — `utf8TermList` includes `dumb`, for one.

The comment's reasoning ("every ENiGMA½ session is drawn with it") is a fair description of how ENiGMA behaves in practice, so this may well be deliberate. But the spec frames lines 8 and 9 as "whether ANSI and RIP respectively are enabled and available to this door session," and that's a question ENiGMA can already answer. Fine to leave as-is if it's intentional — worth saying so in the comment, since as written it reads as an assumption rather than a decision.

core/dropfile.js
```diff
@@ -24,35 +26,173 @@ const { mkdirs } = require('fs-extra');
...
-    static normalizeCommType(commType) {
+    static get CommTypes() {
+        const uint = /^(?:0|[1-9][0-9]*)$/;
+        const none = params => '' === params;
+        return {
+            DORINFO: { local: none, serial: none, socket: none },
+            BBSDEV: {
+                local: none,
+                stdio: none,
+                socket: params => uint.test(params),
```
NuSkooler, line 129: **`io: socket` can never produce a valid file, and the workaround is worse than the failure**

ENiGMA's `io: socket` stands up a TCP listener that the door *dials* via `{srvPort}` (`door.js:76`). The spec's `socket` is an *inherited descriptor* — "an inherited, connected, bidirectional socket," with line 3 carrying the native socket value. Those aren't the same mechanism, and the registry has no token for the one ENiGMA offers.

The docs and the test at `dropfile_bbsdev.test.js` both show this is understood, and refusing is the right instinct. But `defaultDropFileCommType()` still returns `'socket'` for `io: socket`, which then always fails `isValidCommParams` — so BBSDEV plus `io: socket` is guaranteed not to start, and the only way past it is to hand-write a `commParams`. That's the part that worries me: the value would be a descriptor number ENiGMA never arranged to share, so the door would be pointed at whatever that fd happens to be in its own process. Node sets `FD_CLOEXEC`, so in practice it's closed rather than something interesting, but the file would be asserting something untrue and the door has no way to tell.

Given the format's own rule that a producer with nothing honest to write should say so, rejecting `socket` outright — with a message naming `uart` or `fossil` as what a QEMU/DOSEMU bridge should use instead — seems more in keeping than accepting a number that can't be right. That'd also make the failure explain itself at the point of misconfiguration.

andy5995:
To do this PR, I cloned https://github.com/RealDeuce/bbsdev.drp and pointed my bot at it.

NuSkooler (commented):
Thanks for this -- the feature is in good shape, and the blast radius is nicely contained to the opt-in `dropFileType: BBSDEV`. Legacy dropfile semantics look preserved exactly, and the docs and WHATSNEW are unusually thorough.

These notes predate your rebase. I've checked them against `f819a262` and they all still apply -- `core/dropfile.js` and `core/abracadabra.js` are byte-identical across the force-push, so nothing here was overtaken by it.

Two I'd like resolved before merge:

* **the line 12 encoding**, and the allowlist note that goes with it -- a correctness bug that produces mojibake, in the field WHATSNEW leads with
* **`io: socket` with BBSDEV** -- a guaranteed non-start whose only workaround is to hand-write a descriptor ENiGMA never arranged to share

The other three are decisions rather than defects. Each closes just as well with a comment recording that the choice was deliberate as with a change -- the ANSI one already says as much itself.

core/dropfile.js
```diff
@@ -24,35 +26,173 @@ const { mkdirs } = require('fs-extra');
...
+
+//
+//  IANA preferred MIME names for the character sets a session can be set to.
+//  Keys are spelled without punctuation and looked up that way, because the
+//  same encoding reaches us under several spellings: iconv-lite's own, a
+//  sysop's |forceOutputEncoding|, and whatever a menu's setClientEncoding
+//  passes (e.g. 'utf-8'). BBSDEV.DRP line 12 names the character set of the
+//  terminal data, not of the drop file, which is always UTF-8.
+//
+const BbsDevEncodingNames = {
```
NuSkooler, line 38: **The allowlist rejects iconv spellings ENiGMA itself accepts, and the door then refuses to launch**

This table is keyed on a normalised form, but `client_term.js:59` validates a session encoding with `iconv.encodingExists()`, and `term.forceOutputEncoding` is a free-form string assigned literally. Those two sets do not agree:

```
437        iconv=true  allowed=false
850        iconv=true  allowed=false
1252       iconv=true  allowed=false
csibm437   iconv=true  allowed=false
win1252    iconv=true  allowed=false
cp437      iconv=true  allowed=true
latin-1    iconv=true  allowed=true
utf8       iconv=true  allowed=true
```

So a sysop who sets `term.forceOutputEncoding: 437` -- a perfectly good iconv alias -- finds every BBSDEV door on the board refusing to start, with a message that reads like a missing feature rather than a spelling the table does not carry. `setClientEncoding()` in `system_menu_method.js` can put the same value there from a menu form.

Normalising through iconv's canonical name before the lookup would close it.

Raising it here because it is the same root as the line 12 note: whichever value line 12 ends up sourced from still has to survive this lookup, so the two are best settled together.

NuSkooler (commented):
One more in the same area, found while re-checking the notes above against the current head.

andy5995:

Rebased onto master, and each note answered in its own commit.

## The two before merge

**Line 12 now states the door's encoding.** `DropFile` takes an `encoding` option, `abracadabra` passes the same `this.config.encoding` it hands `Door` as `exeInfo.encoding`, and both the field and the `bbsDevError()` guard read it -- so the two cannot diverge, which was the second half of your note. A test sets `term.outputEncoding` to `utf8` against an `encoding: cp437` door and asserts `IBM437`, so the mojibake case is now an assertion rather than a comment.

**The allowlist folds the aliases in** rather than refusing them. `437`, `850`, `1252`, `csibm437` and `win1252` all reach their table entry now, by way of a normaliser that strips a `cs` prefix, expands `win` to `windows`, and maps a bare code page number. I checked the rest of your list against iconv while doing it: `8859-1` and `88591` are *not* accepted by `encodingExists()` either, so those stay refused and the two sets now agree.

**`socket` is gone from the BBSDEV registry.** It is not in `CommTypes.BBSDEV` at all, so a hand-written `commParams` cannot reach it -- `normalizeComm` refuses `socket` with `''`, `'5'` and `'007'` alike, and the message names `uart` with the I/O base and IRQ, or `fossil` with the port, as what a QEMU or DOSEMU bridge should say instead. `DOOR32` still reports `socket` as it always has, which a test now pins.

## The three decisions

**The user key** is recorded rather than changed. The comment states the rowid-reuse mechanism, that it satisfies "stable" and not "never reassigned", and why each alternative is worse here. I confirmed your caveat before writing it: the `AccountCreated` reference in `oputil_user.js` is `listUsers()` reading the property, not a write, so an `oputil user add` account really does lack it.

**ANSI** is likewise recorded. The comment now says it is asserted rather than asked, and that reporting `isANSI()` would tell a `dumb`-type session `N` about a session that is about to receive escape sequences.

**The language tag I fixed rather than annotated**, since your note was that a silent invalid tag sits badly beside a hard refusal three lines up. `Intl.getCanonicalLocales()` throws `RangeError` on a malformed tag, which is the registry's own well-formedness test without a regex of mine to keep current, and `bbsDevError()` now refuses `English (US)` the way it refuses an unnameable encoding.

## Two that did not go the way you suggested

**The schema cannot carry a `pattern`.** `MetaFields` in `core/config/schema.js` lists what meta may set on a node -- `type`, `nullable`, `closedKeys`, `openMap`, `value`, `items`, `enum`, `min`, `max`, `description`, `discriminant`, `variants` -- and there is no pattern among them. Surfacing a malformed tag through `oputil config validate` means extending the validator, which seems worth its own change rather than a line here.

**`core/dropfile.js` moved to a late-bound `Config`.** The language test has to push a config, and a load-time capture of `require('./config.js').get` freezes whichever getter was installed first. `message_area.js` carries a comment about exactly this, and `bluewave_mail_packet.js` picked up the same form in #798, so it is consistent with the tree rather than new -- but it is a line you did not ask for.

Also folded in: the comment in `normalizeComm` pointed at `docs/_docs/modding/local-doors-abracadabra.md`, dead since the site migration, and the docs and WHATSNEW both claimed `socket` was in the registry.

NuSkooler (approved):
Very cool, thank you!

PR #799, opened 2026-09-09
refactor(msg): lift the offline export flow out of the QWK module

andy5995:
> direction.

The packet-gathering flow -- areas, progress, cancel, the download queue handoff -- is not QWK's; a second format needs all of it. Moving it to a base class also fixes an export that failed while finishing: the only 'error' handler recorded the failure and the step went on waiting for 'finished', leaving the caller on the export screen until their session timed out.

### DIFF
--- a/core/message_base_offline_export.js
+++ b/core/message_base_offline_export.js
@@ -1,0 +1,598 @@
+/* jslint node: true */
+'use strict';
+
+//  ENiGMA½
+const { MenuModule } = require('./menu_module.js');
+const Message = require('./message.js');
+const { Errors } = require('./enig_error.js');
+const {
+    getMessageAreaByTag,
+    getMessageConferenceByTag,
+    hasMessageConfAndAreaRead,
+    getAllAvailableMessageAreaTags,
+} = require('./message_area.js');
+const FileArea = require('./file_base_area.js');
+const { renderSubstr } = require('./string_util.js');
+const FileEntry = require('./file_entry.js');
+const DownloadQueue = require('./download_queue.js');
+const { getISOTimestampString } = require('./database.js');
+const { safeMoveFile } = require('./file_util.js');
+
+//  deps
+const async = require('async');
+const _ = require('lodash');
+const fse = require('fs-extra');
+const temptmp = require('temptmp');
+const paths = require('path');
+const { randomUUID } = require('crypto');
+const moment = require('moment');
+
+const FormIds = {
+    main: 0,
+};
+
+const MciViewIds = {
+    main: {
+        status: 1,
+        progressBar: 2,
+
+        customRangeStart: 10,
+    },
+};
+
+//
+//  The flow every offline mail packet export shares: gather the areas the
+//  caller asked for, walk the new messages in each into a packet writer,
+//  then hand the finished packet to their download queue.
+//
+//  A format supplies a writer and a few names by overriding the hooks below.
+//  A writer is expected to emit 'ready', 'packet', 'finished', 'error' and
+//  'warning', and to take messages through appendMessage().
+//
+module.exports = class MessageBaseOfflineExport extends MenuModule {
+    constructor(options) {
+        super(options);
+
+        this.config = Object.assign(
+            {},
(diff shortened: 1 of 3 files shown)


andy5995:
~~This depends on #798 being merged first. After merging, the gap to use the export flow will be implemented here.~~

NuSkooler:
798 says it relies on 799, which says it relies on 788 first :D

andy5995:
> 798 says it relies on 799, which says it relies on 788 first :D

Late night vibe-coding, got my mixed messages up. ;)

core/message_base_offline_export.js
```diff
@@ -0,0 +1,598 @@
...
+    defaultExportOptions() {
+        return { archiveFormat: 'application/zip' };
+    }
+
+    //  MCI views the menu must carry; a format whose art has none returns []
+    requiredViewIds() {
+        return [MciViewIds.main.status, MciViewIds.main.progressBar];
+    }
+
+    noResultsMenuName() {
```
NuSkooler, line 98: `_missingHook()` (line 252) makes this a required hook, but its only call site -- the `'NORESULTS' === err.reasonCode` branch at line 174 -- can never fire. Nothing in this flow produces that reasonCode: the no-results path raises `Errors.NothingToDo` and is handled in the waterfall's final handler via `updateStatus()`. `NORESULTS` is only ever constructed in `file_base_list_export.js` and `file_area_list.js`.

Carrying the dead branch and its `:TODO:` across is the right call for a refactor. But promoting it to a *mandatory* hook is new, and it means every future format has to implement something that cannot be reached. Could you either give it a default here, or drop it from `_missingHook()`?

core/message_base_offline_export.js
```diff
@@ -0,0 +1,598 @@
...
+        throw Errors.MissingParam('noResultsMenuName() is required');
+    }
+
+    //
+    //  A format that lists areas in the packet declares them here, before any
+    //  message is gathered -- Blue Wave lists an area that had no new mail so
+    //  a reader can still post into it.
+    //
+    prepareAreaForExport(/*packetWriter, { areaTag, area, conf }*/) {}
```
NuSkooler, line 107: The signature comment advertises `{ areaTag, area, conf }`, but the private-area call site at line 543 passes only `{ areaTag }` -- `area` and `conf` are `undefined` there.

Blue Wave is unaffected since `addArea()` only wants the tag, but the first format that reaches for `area.name` to build an area list will silently get a blank entry for private mail instead of an error. Worth either noting the asymmetry in this comment, or synthesizing a pseudo-area for the private case so the hook has a single shape.

NuSkooler (commented):
Nice extraction -- the shared flow reads cleanly separated from the format-specific bits, and the two writer-failure fixes are real. I stubbed a writer that emits `error` instead of `ready`, and one that emits it instead of `finished`, and drove `_performExport` on both branches:

```
--- master ---
mode=init    callbackFired=false  idleMonitor=["stop"]
mode=finish  callbackFired=false  idleMonitor=["stop"]
--- this branch ---
mode=init    callbackFired=true   idleMonitor=["stop","start"]
mode=finish  callbackFired=true   idleMonitor=["stop","start"]
```

Worth noting the `init()` case wedges on master too, not just the finish case the description mentions -- and because `stopIdleMonitor()` clears the interval outright while the only `startIdleMonitor()` sits in the handler that never runs, there is no idle timeout left to rescue the node. It hangs until the carrier drops, and `temptmp.cleanup()` leaks the packet dir alongside it. Good catch, and the tests cover both paths.

Two small asks on the new hook API inline. Neither blocks -- they are just cheaper to settle here than across three formats later.

NuSkooler (approved):
Verified the extraction against master line by line -- behaviour-preserving, and I confirmed both writer-failure hangs empirically (a stubbed writer erroring on init() and on finish() never calls back on master and leaves the idle monitor stopped; this branch recovers both). Hook API tidied up in 2cd53468, and the private-area test genuinely resolves the area rather than asserting a shape. Thanks Andy!

PR #801, opened 2026-09-09
build(docker): install from the lockfile so an image is reproducible

andy5995:
> direction.

package-lock.json is tracked but .dockerignore excluded it, so the image resolved dependencies afresh at build time -- and since the three architectures build separately, one run could publish three different dependency trees. npm ci also fails loudly on a lockfile that has drifted from package.json, where npm install quietly resolves around it.

Dev dependencies are still installed, as npm install did here.


NuSkooler (approved):
Verified the premise: `package-lock.json` is tracked and `.dockerignore` was excluding it, so the image really was resolving dependencies fresh -- and with three architectures building separately, one run could publish three different trees under a single tag.

Checked the thing most likely to bite here, `npm ci` installing strictly from a lockfile that omits other platforms' optional deps: this one is v3 and carries all of them (`@img/sharp-linux-arm64`, `-linux-arm`, and the matching libvips packages), so arm64 and armv7 resolve fine. `npm ci --dry-run` also passes, so the lockfile is in sync with package.json today.

Worth noting the Docker workflow has no `pull_request` trigger, so none of this is exercised until it is on master publishing `:latest`. Might be worth a manual `workflow_dispatch` run off this branch before merging.

PR #802, opened 2026-09-09
fix(docs): clear the Gemfile advisories by moving to Ruby 3.4

NuSkooler:
`docs/Gemfile.lock` carried **22 open Dependabot alerts** across six gems, one of them critical. All six are docs-build-only — none ship in the BBS — so the blast radius is the Jekyll toolchain on ephemeral CI rather than anyone's board. Still worth clearing, not least so the alert list stops hiding anything that matters.

| gem | alerts | reached via |
|---|---|---|
| nokogiri | 11 (1 **critical**, 1 high, 3 medium, 7 low) | html-pipeline |
| activesupport | 3 medium | html-pipeline |
| concurrent-ruby | 3 (1 high, 2 low) | activesupport, i18n, tzinfo |
| webrick | 2 (1 high, 1 medium) | direct |
| addressable | 1 high | jekyll |
| rexml | 1 low | kramdown |

## Five of the six were only stale resolution

`html-pipeline` asks for `nokogiri >= 1.4` and `activesupport >= 2`. Nothing was holding those down but the lockfile itself, so a re-resolve lifts activesupport, addressable, concurrent-ruby, rexml and webrick to patched versions with no constraint changes at all.

## nokogiri is the one that needed a decision

Its fixes land in 1.19, and **1.19 requires Ruby >= 3.2** while the workflow pinned 3.1. That pin already carried the comment *"updated to 3.1+ to satisfy nokogiri >= 1.18"* — this is the same thing recurring.

**Why 3.4 and not 3.3**, and why that pulls jekyll along: `csv` left the default gems in Ruby 3.4, and jekyll 4.2 requires it without declaring it, so jekyll 4.2 cannot even boot there — it dies on `cannot load such file -- csv`. Ruby 3.3 *would* have cleared all 22 alerts with no Gemfile change at all, and I built and verified that too. But jekyll 4.2.2 dates from 2022 and already emits the `csv will no longer be part of the default gems` warning, so that route just defers the same break to whoever next raises Ruby. jekyll >= 4.3 declares `csv`, so the pin moves to `~> 4.4` and 4.4.1 resolves.

If you'd rather take the smaller step, the Ruby 3.3 variant is a one-line change and I have it ready.

## jemoji stays

Worth stating, since it's what drags in html-pipeline and therefore 14 of the 22 alerts. The docs lean on it hard — ~340 shortcodes, mostly `:-1:`, `:information_source:`, `:+1:` and `:bulb:` — and jemoji 0.13.0 still wants `html-pipeline ~> 2.2` regardless, so there was nothing to gain by moving or dropping it.

## Testing

The site was built on both the old and new stacks and the output **compared page by page**, not eyeballed. 151 files either way. Everything but six pages and the stylesheet differs only in the generator version string, the build timestamp, and JSON-LD key order (json 2 → 3).

Two engines did change underneath, so the remainder is worth spelling out:

- **SassC → Dart Sass** (via jekyll-sass-converter 3). The only semantic change in the generated CSS is `black` written as `rgb(0, 0, 0)`. The rest of the byte difference is expanded rather than compact formatting.
- **rouge 3.30 → 4.7** retokenizes six pages, and does it better: `1.1` was being split into an integer plus the string `.1` and is now a single float token, and two JS identifiers move `nx` → `nf`. I checked the theme's stylesheet — `mi` and `mf` are both styled, `nx` and `nf` are both unstyled — so nothing renders differently.

Also confirmed: emoji still render (`:warning:` → `emoji/unicode/26a0.png`), the lockfile is stable across a second `bundle install`, and `BUNDLED WITH` stays on 2.x. A default `gem install bundler` would have stamped 4.0.20 and tripped a newer-lockfile warning in CI, so it was pinned to `~> 2.6` deliberately.

## What I could not test

I built with plain bundler in a `ruby:3.4-slim` container, **not** through `limjh16/jekyll-action-ts@v2`, which is what the workflow actually runs. It should simply use the Ruby that `setup-ruby` installed, but that step is unverified here — and unlike a Docker build, a failure shows up on the public docs site. Worth a look at the Pages deploy after merge.


docs/Gemfile
```diff
@@ -8,7 +8,7 @@ source "https://rubygems.org"
 #
 # This will help ensure the proper Jekyll version is running.
 # Happy Jekylling!
-gem "jekyll", "~> 4.2.1"
+gem "jekyll", "~> 4.4"
```
andy5995, line 11: I wonder if you should replace this with the github-pages gem. I don't remember well enough when to do that though. In the case of your site, I won't claim to be certain. But I have this on one of my sites:

```
# This will help ensure the proper Jekyll version is running.
# Happy Jekylling!
#gem "jekyll", "~> 3.7.4"

# This is the default theme for new Jekyll sites. You may change this to anything you like.
# gem "minima", "~> 2.0"

# If you want to use GitHub Pages, remove the "gem "jekyll"" above and
# uncomment the line below. To upgrade, run `bundle update github-pages`.
gem "github-pages", group: :jekyll_plugins
```

The comments were created when I created the Gemfile as I recall. I guess my suggestion is that you might wanna ask your LLM about it, or do a search, unless you already know that what you have is correct. I know your site is working, but it might help prevent future breaks?

andy5995: > Replying as Claude (Opus 5), an LLM made by Anthropic, at andy5995's direction.

I checked the suggestion above and it does not apply here: this workflow builds Jekyll itself (`jekyll-action-ts`) and pushes to `gh-pages`, so that Gemfile is yours rather than Pages'. The `github-pages` gem pins Jekyll 3.10.x and would undo this bump; it is the right advice only for a site GitHub builds from source. What you have is correct.

NuSkooler:
Superseded by #803, which replaced the Jekyll site with Astro + Starlight and deleted the toolchain this was repairing.

All three files this PR touches are gone from `master`:

- `docs/Gemfile`
- `docs/Gemfile.lock`
- `.github/workflows/jekyll.yml`

And the outcome it was chasing arrived anyway — every rubygems advisory is now closed as `fixed`, because the lockfile that carried them no longer exists. Dependabot currently shows 28 open alerts, all npm, none rubygems.

Worth recording that the analysis here was not wasted: the Ruby 3.4 / jekyll 4.4 investigation is exactly why the replacement did not try to keep a Ruby toolchain alive. The `csv`-left-the-default-gems problem you traced would have recurred on the next Ruby bump.

One loose end from that era was still live until today and is now fixed in #805: `.vscode/tasks.json` still ran `cd docs && bundle exec jekyll serve`, and `.devcontainer`'s `postCreateCommand` still did `cd docs && bundle install` — so every dev container build was failing on a directory that no longer existed.

PR #808, opened 2026-09-11
feat(msg): post a new message from a message list

### LINKED ISSUE #213: Message List: Post new
**Feature**:
Ability to post a new message from the message list itself

Using the menu or `m` hotkey by default.

See also: #29

andy5995:
Fixes #213


core/msg_list.js
```diff
@@ -133,6 +133,34 @@ exports.getModule = class MessageListModule extends (
...
+                    formData,
+                    'value.messageIndex',
+                    formData.value.message
+                );
+
+                const areaTag = _.get(this.config, ['messageList', messageIndex])
+                    ? this.getSelectedAreaTag(messageIndex)
+                    : this.config.messageAreaTag;
+
+                return this.gotoMenu(
```
NuSkooler, line 166: The cursor is lost coming back from here.

`selectMessage` twenty lines up sets `this.initialFocusIndex` before it navigates (line 83) precisely so the list restores where the caller was: `menu_stack.goto()` snapshots it via `getSaveState()`, and `prev()` feeds it back through `restoreSavedState()` into `setFocusItemIndex()`.

`postNewMessage` computes the same `messageIndex` but never assigns it, so the value that gets saved is whatever `mciReady` picked as the first unread -- or `undefined`.

Scroll to message 250 of 400, press <kbd>P</kbd>, write the post (or escape out, which also lands in `prevMenu`), and you come back at the top of the list. Every time.

```js
this.initialFocusIndex = messageIndex;
```

before the `gotoMenu` restores parity with `selectMessage`.

WHATSNEW.md
```diff
@@ -3,6 +3,8 @@ This document attempts to track **major** changes and additions in ENiGMA½. For
 
 ## 0.5.1-beta
 
+* **Post from a message list** ([#213](https://github.com/NuSkooler/enigma-bbs/issues/213)) — <kbd>P</kbd> from the message list or the personal message list starts a new message, without going back to the message menu first. The post goes to the area the highlighted message is in, so it does the right thing on a list that spans areas.
```
NuSkooler: Two things on this entry.

It conflicts with master -- #798 landed a Blue Wave bullet at the top of this same section. Keep both. Nothing else in the PR conflicts; I ran three-way merges and `core/msg_list.js` and the menu template both merge clean.

More substantively: this announces <kbd>P</kbd> unconditionally, but `misc/menu_templates/*.in.hjson` only seeds **new** installs. Every existing board's generated `menu.hjson` keeps its current `actionKeys` and never gains the key, so as written this reads as a feature that silently doesn't appear.

`website/src/content/docs/admin/upgrade-notes.md` has a live "0.5.0-beta to 0.5.1-beta" section already carrying menu-edit instructions for exactly this case. Worth an entry there showing the `actionKeys` block to add to `messageBaseMessageList` and `messageBaseMyMessagesList`, with WHATSNEW pointing at it. (#798's <kbd>B</kbd> key has the same gap -- that one's on me to sort separately.)

misc/menu_templates/message_base.in.hjson
```diff
@@ -308,7 +308,16 @@
...
                     }
-                    actionKeys: @reference:common.quitToPrev
+                    actionKeys: [
+                        {
+                            keys: [ "escape", "q", "shift + q" ]
+                            action: @systemMethod:prevMenu
+                        }
+                        {
+                            keys: [ "p", "shift + p" ]
+                            action: @method:postNewMessage
```
NuSkooler, line 323: `menuNewPost` is honoured in the code (`msg_list.js:159`, `this.config.menuNewPost || 'messageBaseNewPost'`) but never surfaces in the shipped template, while its sibling `menuViewPost` is set in this menu's `config` block just above (line 292). A sysop reading their generated `menu.hjson` sees one override and has no way to know the other exists.

Worth adding `menuNewPost: messageBaseNewPost` alongside it, here and in `messageBaseMyMessagesList`.

core/msg_list.js
```diff
@@ -133,6 +133,34 @@ exports.getModule = class MessageListModule extends (
                     return cb(null);
                 }
             },
+            //
+            //  Posting from a list targets the area the focused message is
+            //  in rather than the caller's current area, since a personal or
+            //  search list spans areas. The post module checks write access
```
NuSkooler: The comment says "a personal or search list", but the search results list (`messageBaseSearchResultsMessageList`) and the new-scan list (`newScanMessageList`) are both `msg_list` and both span areas, and neither gets the key here.

The reasoning is right for the two menus it does cover -- either drop "or search" so the comment matches what shipped, or extend the key to those two if you meant to include them.

NuSkooler (commented):
Nice one. The ACS story here is genuinely correct rather than merely asserted: `postNewMessage` hands the tag to `messageBaseNewPost` -> `msg_area_post_fse`, which checks `hasMessageConfAndAreaWrite` before anything else in `initSequence`, and that helper checks both conf and area write. I traced it rather than taking the comment's word for it. Per-row area targeting is the right call for a list that spans areas, and the personal list can't route a post into private mail because `my_messages` builds its filter with no `privateTagUserId`, so `findMessages` takes the explicit exclude-Private branch.

`P` is the right key -- `m` is already `markAllRead` on `newScanMessageList`, so don't worry about #213 naming it.

Two things before merge, inline, then two smaller ones.

One follow-up bigger than this PR: `@reference:` is whole-value substitution only, so inlining the quit bindings detaches both menus from `common.quitToPrev` -- a later change to the shared bindings won't reach them. You're consistent with `newScanMessageList`, which already inlines the same literal for the same reason, so it isn't a new divergence. But `@reference:` does resolve inside array elements, so adding a `common.quitToPrevEntry` *object* beside the existing array would let all three sites extend rather than replace. Worth doing across all three separately.

NuSkooler (approved):
🚀

Issue #108, opened 2017-03-01
Passing a message from point system to uplink


ghost:
First of all, I'm not even sure if enigma can do this or not...

I set up a WWIV fsxnet point node off my enigma 1/2 BBS. I added it as an "uplink" and it's importing packets fine from the WWIV bbs, but it's not then sending the messages on to the agency.

I'm not sure if this works in enigma, or if it's a problem with the packets from WWIV.

NuSkooler:
I have not attempted any of this yet. If you can help me understand what exactly is required, I can tell you how to do it and/or start adding the functionality in.

Currently the system will not re-export any messages it imported (e.g. from FTN). Is it simply a matter of allowing it to re-export them if they are targeted at additional uplink(s)?

apamment:
Sorry, I don't know enough about it. Maybe streaps knows?

streaps:
Re-exporting the messages would be one way to do it. I think most tossers forward/route echomail on-the-fly when they toss (import) the PKTs.

NuSkooler:
> Re-exporting the messages would be one way to do it. I think most tossers forward/route echomail on-the-fly when they toss (import) the PKTs.

I'm not sure of the implementation details, but basically: Drop them in the appropriate outbound area(s) & update SEEN-BY/etc.

streaps:
and optionally ^aPATH (FTS-4)

for Netmail: (^aINTL), ^aFMPT, ^aTOPT (FTS-1) and ^aVia (FTS-4009)

NuSkooler:
Closing as a duplicate of #748, which describes this same gap with a working implementation behind it.

For the record, since this thread asked the right questions back in 2017 and they now have answers:

- @streaps' point about re-exporting vs. forwarding on toss: #748 takes the re-export route (a second scan pass alongside `performEchoMailExport`), which is what this thread converged on.
- The `^aPATH` concern raised here is already handled. `core/scanner_tossers/ftn_bso.js` updates both SEEN-BY and PATH for any EchoMail message on export, keyed on `areaTag` rather than on origin, so a relayed import picks both up through the shared path. MSGID is likewise preserved rather than regenerated, which is what you want for a relay.

The actual blocker turned out to be narrower than anything discussed here: the existing export scan excludes *any* message carrying a `state_flags0` meta row, and import always sets one. See #748 for the detail.

PR #817, opened 2026-09-12
feat(ftn): relay imported EchoMail to secondary uplinks (#748)

### LINKED ISSUE #748: relay imported EchoMail to secondary configured uplinks (points / sub-hubs)
## This is a feature request, not a bug report

Everything below describes something ENiGMA½ **does not currently do at
all** — not a regression, not broken behavior, just an absent capability.
I want to flag that distinction clearly up front, because the fix I'm
proposing is a genuinely new code path with real design decisions in it
(watermarking strategy, SEEN-BY handling for points specifically, batching
strategy), not a small patch to existing logic. I'd expect a maintainer to
want to weigh in on the approach — possibly wanting it opt-in per area,
possibly wanting the SEEN-BY/point handling done differently, possibly
wanting it structured differently entirely. What's here is a working
implementation I've been running against a real point for several days, not
a final proposed design.

## The gap

`core/scanner_tossers/ftn_bso.js`'s existing `performEchoMailExport` — the
only mechanism that currently forwards EchoMail to configured uplinks —
explicitly excludes any message carrying a `state_flags0` meta row:

```sql
SELECT message_id, message_uuid FROM message m
WHERE area_tag = ? AND message_id > ? AND
    (SELECT COUNT(message_id) FROM message_meta
     WHERE message_id = m.message_id AND meta_category = 'System'
       AND meta_name = 'state_flags0') = 0
```

Import always sets `state_flags0 = Imported` on every message it brings in
(`core/scanner_tossers/ftn_bso.js`, import path:
`message.meta.System.state_flags0 = Message.StateFlags0.Imported.toString();`).
Since that condition excludes *any* message with a `state_flags0` row at
all — not a specific value — every imported message is permanently
ineligible for this export scan, regardless of how many uplinks are
configured on the area.

**Net effect: if you configure a second uplink on an echo area — a
downstream point, or in principle a second peer hub — it will only ever
receive messages that were composed locally on that specific BBS by a user
at a terminal. It will never receive echomail that flowed in from the
area's own upstream source.** For a point (whose entire purpose is usually
to read the BBS's live FTN traffic), this makes multi-uplink configuration
essentially non-functional beyond local chatter.

I don't think this has been exercised much before, because in the common
case a BBS's *own* uplinks are all outbound-dialed peers, and a point is the
first scenario where "answering-side, needs mail forwarded from elsewhere"
actually matters end-to-end.

## Proposed approach

A second, independent scan pass — `performEchoMailRelayExport` — added
alongside the existing `performEchoMailExport`, in the same `performExport`
dispatch array:

```js
async.eachSeries(
    ['EchoMail', 'EchoMailRelay', 'NetMail'],   // new entry
    ...
);
```

Key design points, each of which I'd want a maintainer's opinion on:

**1. Separate watermark, not reused from the local-post scan.**
Tracked via a distinct `scan_toss` value (`ftn_bso_relay`) in the existing
`message_area_last_scan` table, so it never interferes with the existing
local-post export watermark. Selects specifically
`state_flags0 = Imported` (not just "any row"), so a message that's already
been relayed (and thus additionally carries `state_flags0 = Exported`,
alongside the still-present `Imported` row — see the companion
`persistMetaValue`/`updateMetaValue` issue for why both rows coexist) isn't
re-selected by *this* scan's own criteria either, though the watermark is
the real, primary safeguard against re-processing.

**2. SEEN-BY-aware target filtering, with an explicit carve-out for points.**
For each eligible message, compute which of the area's configured uplinks
haven't already seen it (via the message's own `ftn_seen_by` meta,
parsed with the existing `ftnUtil.parseAbbreviatedNetNodeList`), and only
relay to those. This is necessary to avoid naive fan-out becoming an
echo-loop generator for any *genuine* secondary peer relationship.

However, SEEN-BY has no point-number component (FTS-1027) — a point's
net/node is identical to its own boss's, and the upstream sender routinely
adds the boss's own net/node to SEEN-BY as a matter of ordinary routing
courtesy before the message ever reaches the boss. That makes a plain
SEEN-BY check indistinguishable from "have I myself already seen this,"
which is always true by the time a message has been imported — so a point
can *never* pass a naive SEEN-BY check, regardless of whether it's
genuinely already seen anything. The implementation special-cases this:
any uplink that is a point of the system's *own* configured local address
is always relayed to, unconditionally; only genuine separate peer uplinks
go through the real SEEN-BY loop-check.

```js
if (
    addr.point &&
    localAddress &&
    localAddress.net === addr.net &&
    localAddress.node === addr.node
) {
    return true;   // always relay to our own points
}
return !alreadySeen.some(
    seen => seen.net === addr.net && seen.node === addr.node
);
```

**3. Batch across areas, not per-area or per-message.**
This one matters for correctness, not just efficiency.
`exportMessagesByUuid` already batches multiple messages into shared
packets/archives correctly (it loads each message individually by UUID and
uses its own stored `area_tag` for the `AREA:` kludge — confirmed via
source, so cross-area batching doesn't misattribute anything). But the
per-destination bundle filename only has 36 possible daily slots
(`getOutgoingBundleFileName`: `<2-letter day><0-9,a-z>`, e.g. `mo0`–`moz` on
a Monday). Calling the export function once per area (or worse, once per
message) burns through those slots fast under any real volume, and once
they're exhausted for the day, later export attempts for that destination
silently fail to acquire a filename at all. The proposed pass collects
*all* eligible messages across *all* configured areas that share the same
`(network, uplink-set)` combination first, then makes exactly one export
call per unique combination — typically one call total for a single-point
setup, regardless of how many areas are wired to it.

## What's NOT included / open questions for a maintainer

- No config toggle to opt an area in or out of relay specifically — any
  area with 2+ configured uplinks is currently eligible. A per-area or
  per-network `enableRelay` flag might be preferable.
- No handling for the case of two genuine peer hubs (not points) both
  configured on the same area — the SEEN-BY loop-check path exists for
  this, but I haven't had a real second-peer scenario to test it against;
  only the point carve-out is battle-tested.
- Relies on the companion `persistMetaValue` → `updateMetaValue` fix (see
  separate issue) to actually produce output reliably — without it, this
  feature silently produces `state_flags0`/`MSGID` constraint failures on
  every relayed message and drops content with no visible error.
- A related, separate finding while building this: `bso_spool.js`'s
  answering-side session handler (used when the point/peer connects in) had
  its own pre-existing bug preventing queued mail from being offered during
  that session at all (see separate issue) — worth being aware this
  feature depends on that too for end-to-end delivery to actually work for
  points specifically, though it isn't part of this feature's own code.

## Testing performed

Running for several days against a live FidoNet point (a point address
sharing net/node with the boss's own address, several areas configured with
the point as a second uplink alongside the area's normal upstream hub).
Confirmed via the point's own tosser log that messages from multiple
distinct areas, batched into a single archive, import correctly and
separately into their respective local areas on the point side — no
cross-area misattribution.

NuSkooler:
Closes #748. Built from @jriethoven's analysis and field notes on that issue — the design decisions they raised are answered below, with two additions.

## The gap

`performEchoMailExport` excludes any message carrying a `System/state_flags0` row, and import always writes one. The condition tests for a row's *presence*, not a value, so every imported message is permanently ineligible regardless of how many uplinks an area has.

Configure a second uplink on an echo — a downstream point, or a sub-hub — and it receives only messages composed by a user at a terminal on this system. It never sees the echomail that flowed in from the area's own source, which for a point is the entire reason it exists.

## What this adds

A second scan pass, `performEchoMailRelayExport`, in the same dispatch.

Most of what it needs already existed: `exportEchoMailMessagesToUplinks` has always looped every configured uplink, and `prepareMessage` already updates SEEN-BY and `^aPATH` for any exported EchoMail regardless of origin, while preserving the original MSGID rather than regenerating it. (That last point answers @streaps' 2017 concern on #108.) What was missing was a way for imported mail to reach it.

**Opt-in per area** via `relay: true`. Relaying is visible to the rest of the network; a board that configured a second uplink for another reason should not silently start feeding it. This was listed as an open question on the issue — I think opt-in is the only safe default.

## Where a message goes

Three rules, in order:

1. **Never back to the system that sent it.** Read from the packet header origin recorded at import (`ftn_orig_*`), which export does not overwrite in the database. This is the addition I'd most want reviewed: it's exact, and it holds when a sender omits itself from SEEN-BY, which the SEEN-BY rule alone does not.
2. **Always to our own points.** SEEN-BY has no point component (FTS-1027) — a point's net/node are its boss's, and an upstream sender routinely lists the boss before the message arrives, so a plain SEEN-BY test reads "already seen" for every point, every time.
3. **Otherwise only when the uplink is not already in SEEN-BY** — the ordinary loop guard for a genuine peer.

## Two things worth a close look

**The relay watermark is seeded, not started at zero.** `getAreaLastScanId` returns 0 both for "never scanned" and "nothing seen yet". A relay scan beginning at 0 selects every imported message in the area — on an established board, the entire message base offered to a link that never asked for it, and enough bundles to exhaust a destination's 36 daily names (`getOutgoingBundleFileName`) and stall. On first sight of an area we record where it is and export nothing. This wasn't in the issue and is the failure I'd most expect in the field.

**Grouped by `(network, uplink)` across areas**, not per area — same bundle-name reason, one export call per destination per run. Safe across areas because `exportMessagesByUuid` loads each message individually and the AREA kludge comes from the message's own `area_tag`; only the network must match, since that supplies the local address.

## API change

`exportEchoMailMessagesToUplinks` now takes its uplink list as a parameter, since relay sends a given message to a subset. The SEEN-BY line still names *every* configured uplink — `prepareMessage` builds it from config — which is correct FTN practice and helps a downlink we skipped avoid sending it back to us.

## Testing

Nine new tests (`test/ftn_echomail_relay.test.js`); suite goes 2528 → 2537, no failures.

Each test was checked against a mutation of the behaviour it covers — removing the seeding, the origin check, the point carve-out, the SEEN-BY guard, the opt-in gate, or the failure hold-back each fails its own test and no others.

That pass found a real problem: **the first version of the origin test passed with the origin check removed entirely.** Its fixture named the hub in SEEN-BY, so rule 3 covered the case and rule 1 was never exercised. It is now written with no SEEN-BY line at all, which is the case that actually distinguishes them.

## Still open

- **Genuine peer-to-peer relay** (another hub, not a point) rests on rule 3 and is far less exercised than the point path — flagged in the docs.
- Per-*network* opt-in, rather than per-area, if a hub with 40 echoes finds the per-area flag tedious.

### DIFF
--- a/core/ftn_mail_packet.js
+++ b/core/ftn_mail_packet.js
@@ -771,6 +771,24 @@ function Packet(options) {
             ftn_msg_dest_net: msgData.ftn_msg_dest_net,
         };
 
+        //
+        //  Record the origin point when the packet is a type that actually has
+        //  one. Without this, mail from our own point reads as coming from our
+        //  own address, and anything reasoning about where a message came from
+        //  -- EchoMail relay, for one -- cannot tell the two apart.
+        //
+        //  Gated on version rather than written unconditionally: origPoint is
+        //  a fill field in a plain type-2 packet, so a stray non-zero value
+        //  there would make a message look like it came from a point that does
+        //  not exist, and the origin would stop being recognised at all.
+        //
+        if (
+            ('2+' === header.version || '2.2' === header.version) &&
+            header.origPoint > 0
+        ) {
+            msg.meta.FtnProperty.ftn_orig_point = header.origPoint;
+        }
+
         self.processMessageBody(msgData.message, messageBodyData => {
             msg.message = messageBodyData.message;
             msg.meta.FtnKludge = messageBodyData.kludgeLines;
--- a/core/message.js
+++ b/core/message.js
@@ -766,6 +766,59 @@ module.exports = class Message {
         }
     }
 
+    //
+    //  OR |bit| into this message's state_flags0, as one row.
+    //
+    //  updateMetaValue() cannot do this. message_meta's UNIQUE key includes
+    //  meta_value (see database.js), so REPLACE INTO with a value that differs
+    //  from what is stored has nothing to collide with and inserts a *second*
+    //  row -- leaving a bitmask field holding '1' and '2' rather than '3', and
+    //  reloading as an array rather than a value.
+    //
+    //  That was unreachable while nothing ever exported a message it had
+    //  imported. EchoMail relay does exactly that.
+    //
+    setStateFlags0Bit(bit, cb) {
+        const name = MessageConst.SystemMetaNames.StateFlags0;
+        try {
+            msgDb.transaction(() => {
+                const rows = msgDb
+                    .prepare(
+                        `SELECT meta_value FROM message_meta
+                        WHERE message_id = ? AND meta_category = 'System'
+                          AND meta_name = ?;`
+                    )
+                    .all(this.messageId, name);
+
+                const flags = rows.reduce(
+                    (acc, row) => acc | (parseInt(row.meta_value, 10) || 0),
+                    bit
+                );
+
+                msgDb
+                    .prepare(
+                        `DELETE FROM message_meta
+                    WHERE message_id = ? AND meta_category = 'System'
+                      AND meta_name = ?;`
+                    )
+                    .run(this.messageId, name);
+
+                msgDb
+                    .prepare(
+                        `INSERT INTO message_meta (message_id, meta_category, meta_name, meta_value)
+                    VALUES (?, 'System', ?, ?);`
+                    )
+                    .run(this.messageId, name, flags.toString());
+
+                this.meta.System = this.meta.System || {};
+                this.meta.System[name] = flags.toString();
+            })();
+            return cb(null);
+        } catch (err) {
+            return cb(err);
+        }
+    }
+
     updateMetaValue(category, name, value, cb) {
         try {
             if (!Array.isArray(value)) {
--- a/core/scanner_tossers/ftn_bso.js
+++ b/core/scanner_tossers/ftn_bso.js
@@ -822,7 +822,7 @@ function FTNMessageScanTossModule() {
     //  Mail imported from here on is relayed; anything already in the base is
     //  treated as history.
     //
-    this.seedRelayScanId = function (areaTag, cb) {
+    this.seedRelayScanId = function (areaTag, scanToss, cb) {
         let maxId = 0;
         try {
             const row = msgDb
@@ -836,12 +836,10 @@ function FTNMessageScanTossModule() {
         }
 
         Log.info(
-            { areaTag, lastScanId: maxId },
-            'EchoMail relay: first scan of this area; starting from current mail'
+            { areaTag, scanToss, lastScanId: maxId },
+            'EchoMail relay: first scan of this area for this uplink; starting from current mail'
         );
-        return this.setAreaLastScanId(areaTag, ScanToss.Relay, maxId, err =>
-            cb(err, maxId)
-        );
+        return this.setAreaLastScanId(areaTag, scanToss, maxId, err => cb(err, maxId));
     };
 
     //
@@ -1049,10 +1047,13 @@ function FTNMessageScanTossModule() {
                             );
                         },
                         function storeStateFlags0Meta(callback) {
-                            message.updateMetaValue(
-                                'System',
-                                'state_flags0',
-                                Message.StateFlags0.Exported.toString(),
+                            //  OR the bit in rather than writing the value:
+                            //  a relayed message already carries Imported, and
+                            //  message_meta's UNIQUE key includes meta_value,
+                            //  so a plain write leaves two rows instead of a
+                            //  combined one.
+                            message.setStateFlags0Bit(
+                                Message.StateFlags0.Exported,
                                 err => {
                                     callback(err);
                                 }
@@ -1414,10 +1415,11 @@ function FTNMessageScanTossModule() {
                             );
                         },
                         function storeStateFlags0Meta(callback) {
-                            return message.updateMetaValue(
-                                'System',
-                                'state_flags0',
-                                Message.StateFlags0.Exported.toString(),
+                            //  See the EchoMail export path: OR the bit in so a
+                            //  message that already carries another flag ends up
+                            //  with one combined value rather than two rows.
+                            return message.setStateFlags0Bit(
+                                Message.StateFlags0.Exported,
                                 callback
                             );
                         },
@@ -1679,7 +1681,15 @@ function FTNMessageScanTossModule() {
 
                                     fse.move(oldPath, newPath, err => {
                                         if (err) {
-                                            Log.warn(
+                                            //  The packet is still in the temp
+                                            //  area, so this export did not
+                                            //  happen. Report it: the caller
+                                            //  advances the area's last scan ID
+                                            //  on a clean return, and doing that
+                                            //  here is how a disk-full or
+                                            //  permission error turns into mail
+                                            //  nobody ever sees again. See #818.
+                                            Log.error(
                                                 {
                                                     oldPath: oldPath,
                                                     newPath: newPath,
@@ -1688,7 +1698,7 @@ function FTNMessageScanTossModule() {
                                                 'Failed moving temporary outbound file!'
                                             );
 
-                                            return nextFile();
+                                            return nextFile(err);
                                         }
 
                                         const flowFilePath = self.getOutgoingFlowFileName(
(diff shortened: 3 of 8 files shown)


NuSkooler:
Ran an independent review over this branch. It found three tests that passed for the wrong reason and two defects outside the diff. Fixes for the first are pushed as `af3aa765`; the rest are recorded below and in #818.

## Corrections to what this PR claimed

**The origin rule does not stop our own point being sent its own mail back.** I wrote that it did, both here and on #748. It does not, and the reason is upstream of this code: `ftn_mail_packet.js` records `ftn_orig_node` and `ftn_orig_network` from the packet header and *nothing else about the origin* — no `ftn_orig_zone`, no `ftn_orig_point`, even for a type 2+ packet that carried them. So a message from our point reads as coming from our own address with no point, the origin rule cannot tell it apart from us, and the point carve-out then relays it back. The receiving tosser drops it on MSGID, so it is wasted traffic on a point link rather than a loop — but the claim was wrong.

Now covered by a test named for what it actually asserts, which will start failing if the importer ever begins recording the point. Fixing it properly means persisting `ftn_orig_point` at import, which needs packet-version care (`origPoint` is a fill field in a plain type-2 packet, so writing it unconditionally would risk *breaking* the origin rule in the common case). That belongs in its own change, not bolted onto this one.

**The test fixture wrote a zone that no real import writes**, so every test drove `isSameFtnSystem`'s strict-zone branch and none drove the zoneless branch production always takes. Mutating that fallback changed no test result. Fixed, with direct coverage added.

**`advances past messages every uplink had already seen` never saw that case.** It ran against an area whose uplinks include a point; points are relayed to unconditionally, so mail was always going out. It asserted only the watermark and passed with the SEEN-BY filter disabled outright. It now runs against a relay area of ordinary peers and asserts nothing was sent.

Seven mutations, one per behaviour — each now fails its own test. Before `af3aa765`, three of them failed none.

## Two defects outside this diff

**#818 — a failed move into outbound is logged, then reported as success.** This is the serious one. `moveFilesToOutgoing` calls its iteratee callback with no error on `fse.move` failure, so the export callback lies and the scan ID advances past mail sitting orphaned in the temp directory. It is pre-existing and affects the ordinary export scan identically, but it means **this PR's "the watermark is held back on failure" guarantee is only true for the failure mode the test happens to use** (`mkdirs`, which does propagate). Disk full, quota, and permission-denied-on-an-existing-directory all take the swallowed path. Filed separately with a reproduction.

**`state_flags0` becomes two rows rather than a combined bitmask.** `updateMetaValue` uses `REPLACE INTO` against a table whose UNIQUE key includes `meta_value`, so writing a different value inserts rather than replaces. A relayed message ends up with both `"1"` (Imported) and `"2"` (Exported) instead of `"3"`. This was unreachable before — nothing ever exported a message that already carried a `state_flags0` — so **this PR is the first thing to hit it.** No current consumer is affected (the only bitmask reader is scoped to `private_mail`), but `message.meta.System.state_flags0` now reloads as an array for such messages, and the field is designed as a bitmask. Worth deciding whether to fix here or track separately — I have not touched it.

## Known limitation, unchanged

Failure tracking is per-area, not per-(message, uplink). If one uplink of an area fails, the whole area's watermark is held back, so messages already delivered to the healthy uplinks are sent again next pass. Deduped on MSGID at the far end. Fixing it properly means per-destination tracking, which is a larger change than this PR should carry.

NuSkooler:
Every finding from the review is now fixed rather than recorded — `f9dc7358`. Nothing is left as a known gap.

| Finding | Status |
|---|---|
| Fixture wrote a zone no real import writes | Fixed, `af3aa765` |
| `advances past...` test never saw its own case | Fixed, `af3aa765` |
| Origin rule did not cover our own point | **Fixed** — importer now records `ftn_orig_point` |
| #818 outbound failures reported as success | **Fixed**, both sites |
| `state_flags0` two rows instead of a bitmask | **Fixed** — `Message.setStateFlags0Bit` |
| Per-area failure re-sending to healthy uplinks | **Fixed** — watermarks are per destination |

## The two that grew in the fixing

**#818's second site matters more than I first said.** The swallowed `flowFileAppendRefs` error includes a **busy flow file** — an ordinary outcome under concurrent sessions, not an error condition. Swallowed, a node that was merely busy had its mail dropped permanently, because the scan ID advanced anyway. That is far more common than the disk-full case I filed the issue about.

**The per-area watermark was worse than a one-time duplicate.** A failure held the entire area back, so the next pass re-selected a set that only grew and handed all of it to the uplinks that were working — a dead point becoming an ever-larger duplicate feed for everyone else. Watermarks are now per `(area, destination)`.

Measured, not argued: against the old model a broken peer had the healthy point receive the first message **three times** across three passes, the second twice, the third once. Per destination, each arrives once. The test that shows this was run against the previous commit's implementation to confirm it discriminates.

## Two bugs the fixes found in each other

Worth recording, because both were silent:

- Combining `state_flags0` properly broke the relay candidate query, which matched `state_flags0 = '1'` exactly and stopped matching once the flags became `'3'`. It tests the bit now — which is also what the retry after a failed pass depends on. Caught by the retry half of the new #818 test.
- My first version of the dead-uplink test counted flow *references*, which count passes rather than messages, since a pass bundles everything for a destination into one packet. It passed against both implementations. It counts delivered messages now.

## On method

15 mutations, one per behaviour, each failing only its own test. Two escaped on the first sweep and are now covered: the flow-append site of #818, and `ftn_orig_point` (the relay tests insert meta directly, so the importer needed its own tests in `ftn_mail_packet.test.js`).

Suite 2528 → 2549. #818 is closed.

PR #822, opened 2026-09-12
feat(bluewave): read reply packets back in

### LINKED ISSUE #810: Blue Wave: read reply packets back in (.UPL / .NET / .UPI)
Export landed in #798 (55bb48a7): a caller can take a Blue Wave packet away and read it offline. What they cannot do is send anything back.

A reader writes its replies into the packet it downloads, as some combination of:

* `*.UPL` — reply message information (the level 3 form, an alternative to the pair below)
* `*.NET` — NetMail reply information
* `*.UPI` — information for all other reply messages
* `*.REQ` — a list of files to request from the BBS

`INF_HEADER.uses_upl_file` is what advertises this, and `core/bluewave_mail_packet.js` currently writes `0` there deliberately, so readers are told up front that replies are not accepted. That byte flips to non-zero when this is implemented.

Work involved, roughly:

* a `BlueWavePacketReader` alongside the writer, taking the uploaded bundle apart
* routing each reply to an area by its 20 character echotag rather than an area number, which is the whole reason the format carries one
* the usual import guards: ACS on the target area, the uploading user's identity, size and count limits, and a packet whose members disagree with each other
* `orig_zone`/`orig_net`/`orig_node` and the internet kludges (`\001From:`, `\001Message-ID:`, `\001References:`) for areas whose `network_type` is Internet

@andy5995 offered to take this on in #798.

Splitting this out so #119 can close on the export half.

andy5995:

Fixes #810.

Blue Wave mail goes both ways now: a caller downloads a packet, replies to it offline, and uploads the reply packet their reader wrote.

### Reading a reply packet

`BlueWavePacketReader` takes apart a `*.UPL` at level 3, or a `*.UPI` and `*.NET` at level 2, plus the `*.REQ` file request list. The kit is explicit that a door reads the `*.UPL` whenever one is present and the older pair only when it is not, or a reader that wrote both has its replies imported twice. Records are seeked by the lengths the header declares rather than this build's own, so a newer reader may add fields to either structure.

### Importing them

A new `message_base_offline_import` menu module, reached with <kbd>U</kbd> from the message base menu. Nothing asks the caller what format they are uploading: a reply packet names itself from the inside, so the archive is opened and read by whichever reader matches its members. The format is chosen on the way *out*, at the export, which is where a caller already picks one.

Each reply is placed by echotag -- against the tags that caller's own export would have written, so an area they cannot see was never in their packet -- and then checked as though they had typed it at the keyboard:

* write access on the target area, and a user who exists for personal mail;
* the `From` name replaced with their own. The kit tells doors to validate that field and there is nothing to validate it against, since a reader writes whatever name it was configured with;
* a packet built for another login refused, on the name the reader copied out of `INF_HEADER.loginname`;
* a count and length limit, since the packet is written by software on the caller's machine.

A reply that cannot be placed is logged and skipped rather than failing the upload.

### What the export now has to carry

A reply is addressed from the packet it answers, so three things changed on the way out:

* `FTI_REC.msgnum` is the message's own ID rather than a number local to the packet, which is what a reply names on the way back. The field is 16 bits and an ID is not, so an ID past that is written as zero -- the kit's "no number" -- rather than truncated into another message's, and the export warns once when it happens. An import checks the ID against the area before chaining anything onto it.
* `FTI_REC.orig_zone/net/node` carry a netmail message's origin address.
* A message in an Internet area carries the `\001From:`, `\001Message-ID:` and `\001References:` kludges MultiMail lifts out to address and thread a reply. In a FidoNet area it would leave them sitting in the text, so they go only where they are read.
* `uses_upl_file` flips to one. A reader writes replies only when the host says it can process them.

### A QWK fix rides along

The first commit fixes QWK reply mode, which had never worked: it required a `CONTROL.DAT`, which a reply packet does not carry, and looked for the messages under the fixed name `ID.MSG` rather than the `<BBSID>.MSG` a reply packet uses. Conference numbering moves out of the writer into `buildConferenceMap()` so an import can reproduce the numbering an export used, and config access in that module is late bound, as in `message_area.js`, so the numbering can be tested. The same upload entry reads a QWK `*.REP`.

### Verified against a real reader

The unit tests here read bytes these tests wrote, which proves the parser agrees with itself and nothing more. So: a packet from this writer was opened in NoCarrierMail 0.55, a reply written in it, and that reply read back here. Echotag routing, the identity fields, the body and the threading all survived the trip -- the reader displayed `Msg#: 4712` and named it in the reply it wrote.

It caught one bug no test here would have. The Blue Wave kit says each byte of the reader version is "the actually ASCII value **plus** 10"; `bw.cc` writes `vernum[c] -= 10`, and a real packet holds `"&$++"` for 0.55. Decoding it as the kit describes gives control characters. There is now a test built from the bytes that reader actually wrote.

### What this does not do

* A message whose ID is past 65535 exports with no number, so replies to it arrive unthreaded. That is deliberate -- the alternative is threading onto whatever message the truncated number names.
* `*.REQ` entries are read and reported, but nothing acts on a file request yet.
* <kbd>U</kbd> is seeded from `misc/menu_templates/`, so it reaches new installations only; `UPGRADE.md` says what to add to an existing `menu.hjson`. The shipped `MSGMNU.ANS` names the key, a theme of your own needs the line.

### Commits

1. `fix(qwk): read reply packets`
2. `feat(bluewave): read reply packets, and carry what a reply needs`
3. `feat(msg): import replies from an uploaded offline mail packet`
4. `docs: offline mail is two-way`


andy5995:
This PR includes [a80f55e](https://github.com/NuSkooler/enigma-bbs/pull/822/commits/a80f55e95e1b14794eac0245f9caa96911c0978e) which is PR #821 . In theory, it's not dependent on this PR, but this PR is dependent on it. You could merge that separately, or close it and just merge this one, depending on your tastes. :)

andy5995:
I pushed #821 about the same time as you merged. I'll rebase this and sort things out. A moment...

andy5995:

Correcting the comment above: #821 has merged as 2857e8d3, so this PR no longer carries that commit and the link above points at a sha that is no longer part of it. The branch is rebased onto master.

One follow-up to it remains here, "fix(qwk): only take a .MSG as the messages file of a reply packet". Matching any `*.MSG` made detection depend on `readdir` order, so a stray `.MSG` beside a `MESSAGES.DAT` could be enumerated first and have a downloaded packet parsed as a reply. It is now honoured only when the caller has said the packet is a reply, which is what both `oputil` and the import do.

andy5995:
This is ready for review.

core/message_base_offline_import.js
```diff
@@ -0,0 +1,595 @@
...
+            if (err) {
+                return cb(err);
+            }
+
+            const identityError = this._checkPacketUser(packetUser);
+            if (identityError) {
+                return cb(identityError);
+            }
+
+            if (pending.length > this.limits.maxMessages) {
```
NuSkooler: **QWK does not get the early limits Blue Wave does.**

The comment above `Limits` has exactly the right instinct — *"neither the record count nor the length of a message is anything this end should trust"* — and the Blue Wave reader implements it: `maxMessages` is checked against `records.length` before any message file is opened (`core/bluewave_mail_packet.js:1144`), and `maxMessageLength` against `stats.size` before the file is read (`:1192`).

The QWK path does not. `QWKPacketReader` takes neither option, so every message in the `.REP` is parsed and accumulated into `pending[]` first, and only then is the count tested here. `maxMessageLength` is not consulted until `_persistMessage`, by which point the whole body is already in memory.

So a `.REP` carrying 100,000 messages is fully parsed before the 500 limit applies. It is an authenticated, ACS-gated path, so this is a resource concern rather than anything sharper — but the asymmetry does mean the QWK format is not actually getting the protection that comment describes.

The cheapest fix is probably at the shared collection point rather than in the reader:

```js
source.reader.on('message', message => pending.push(message));
```

Counting there and bailing past `maxMessages` would cover both formats and keep the per-format readers simpler.

If the upload or extraction path already bounds this upstream, say so and I will drop it.

NuSkooler (commented):
Read this through against the guards #810 asked for. They are all here, and a couple of the decisions are subtler than they look:

* **Identity is taken from the session, not the packet.** `_persistMessage` overwrites `fromUserName` and calls `setLocalFromUserId` regardless of what `UPL_REC.from` says, with a comment explaining that a reader writes whatever name it was configured with. Right call.
* **`replyToMsgId` is only honoured when the target is in the same area.** That closes cross-area reply linking, which I would have asked about otherwise.
* **Personal mail must resolve to a real user** before it persists, and the name is re-spelled as the account spells it.
* **An unmapped QWK conference emits a warning rather than guessing an area** — that was the failure mode I was most worried about coming into this.

**The map-scope asymmetry between the two formats is correct, and worth keeping.** QWK builds its conference map from `getAllAvailableMessageAreaTags()` with no client — every area — because the numbering is positional and a shorter list would renumber. Blue Wave builds its echotag map from the *caller's* visible areas, which is safe precisely because it routes by name. Side by side these look inconsistent; they are not, and the comment on the QWK side explains why. Worth leaving that comment in if anyone later tries to unify them.

Full suite is green — 2613 passing. `test/bluewave_packet.test.js` run standalone shows failures, but `master` shows the same pattern on that file, so that is the known "only the default mocha order is green" behaviour rather than anything here.

One substantive thing inline, plus a question.

---

**`uses_upl_file` is written unconditionally, and UPGRADE.md is honest that this has a consequence:** every exported packet now tells the reader to write replies, whether or not the sysop has added the <kbd>U</kbd> key to their `menu.hjson`. On an upgraded board that means a caller can compose replies offline and then find there is nowhere to upload them.

The note covers it clearly, and I do not see a clean way for the writer to know whether the menu entry exists — so this may simply be the right trade. Asking because it is the one user-visible rough edge for existing boards, and if there *is* a signal available it would be worth keying off.

On the conference numbering: the `qwk.md` addition already says an automatic number "moves when you add or remove an area, and a reply written against the old numbering would then name a different one", and recommends pinning. That is the mitigation and it is documented, so nothing needed here. The only open question is whether a log warning at export time is worth it for a board running two-way QWK on automatic numbering — I would leave it unless you want it.

core/message_base_offline_import.js
andy5995: > Posted by Claude (Opus 5), an LLM made by Anthropic, at andy5995's direction.

Nothing bounds it upstream, so the concern stands as written. Fixed at the collection point you pointed at, which covers both formats:

```js
source.reader.on('message', message => {
    if (pending.length >= this.limits.maxMessages) { overLimit = true; return; }
    if (message.message.length > this.limits.maxMessageLength) { /* counted, dropped */ return; }
    pending.push(message);
});
```

An oversized body is dropped there as well, so it is not held in `pending[]` until `_persistMessage` looks at it; the now-redundant length check in `_persistMessage` is gone.

The Blue Wave reader keeps its own earlier checks rather than deferring entirely to this. It knows the record count from the `*.UPL` before it opens a single message file, and it sizes each file with `stat` before reading it, so an oversized packet is refused without ever being read. The QWK reader cannot do the same -- it walks the packet in 128 byte blocks and only knows what a message is once it has parsed it -- which is exactly why the shared check is the one that has to exist.

andy5995:

On `uses_upl_file`: there is a signal, and the byte now follows it.

`client.currentTheme.menus` is the caller's merged menu set, so the export can ask whether any menu carries `message_base_offline_import` and write the byte from the answer (82f67887). A board that has not added the key advertises zero, so nobody is invited to write replies they have nowhere to send; add the key and packets start advertising one, with nothing to configure twice. It is per caller rather than per system, which is the right granularity -- the menus a caller reaches are what decides whether they have an upload.

That inverts what the UPGRADE note said, so it and the two doc pages are corrected in 61b1af67: the note now says that leaving the key out breaks nothing, rather than warning about the consequence of it.

On an export-time warning for automatic QWK conference numbering: leaving it out. It would fire on every export from any board that has not pinned conferences, including the many that will never see a reply packet, and the advice it would carry is already on the page next to the setting. If two-way QWK turns out to be common on automatically numbered boards, a warning at *import* time -- when a reply names a conference that no longer maps -- would be the one that fires only when it matters, and that path already emits one.

The map-scope comment is staying where it is.

NuSkooler (approved):
Sweet! Now I need to get Xibalba updated with some art for this work!

andy5995:
> Sweet! Now I need to get Xibalba updated with some art for this work!

Looking forward to it!

PR #825, opened 2026-09-13
fix(doors): preserve DOOR.SYS call times

### LINKED ISSUE #824: Drop files: the call time is written in 12-hour format, and "Time of Last Call" is given this call's time
Two defects in `core/dropfile.js`, found while researching #795 but independent of it — neither needs any decision about time limits, and both can be fixed on their own.

## 1. The call time is 12-hour

`core/dropfile.js:350`:

```js
const timeOfCall = moment(prop[UserProps.LastLoginTs] || moment()).format(
    'hh:mm'
);
```

In moment, **`hh` is 12-hour**; `HH` is 24-hour. Verified:

```
moment('2026-09-12T14:32:00').format('hh:mm')  ->  02:32
moment('2026-09-12T14:32:00').format('HH:mm')  ->  14:32
```

So a caller who logs on at 14:32 is announced to every door as having logged on at **02:32**, with no AM/PM marker to disambiguate. `timeOfCall` feeds DOOR.SYS lines 44 and 45.

Every reference implementation writes 24-hour here — the GAP DOOR.SYS document, Synchronet, Mystic, Renegade and EleBBS all agree, and the format has no room for a meridiem indicator.

**Fix:** `'HH:mm'`.

## 2. "Time of Last Call" is given *this* call's time

`core/dropfile.js:403-404`:

```js
timeOfCall, //  "Time of This Call"
timeOfCall, //  "Time of Last Call                 (hh:mm)"
```

The same value is written to both. Line 45 is supposed to be the *previous* call.

It cannot currently be anything else: `recordLogin()` in `core/user_login.js` overwrites `LastLoginTs` at login, and `core/user_login.js:197` reads the previous value for the login-streak computation and then discards it. There is no stored previous-login timestamp anywhere.

**Fix:** persist the prior value as a `PrevLoginTs` user property during `recordLogin()` — one `setUserStat` beside the existing streak logic — and read that for line 45. A first-time caller has no previous call, so a blank line is the honest value there.

## While in the file

`CALLINFO.BBS` is commented `//  Citadel?`. It is **Wildcat!'s** native drop file — `USERINFO.DAT` on the following line is correctly attributed. Comment only; the format is not implemented.

## Testing

`test/dropfile.test.js` would want a case pinning the 24-hour format across an afternoon timestamp (a morning one passes either way, which is how this survived), and one asserting that lines 44 and 45 differ on a second call.

sjh9714:
## Summary

Fixes #824.

Write the current and previous login times to their respective DOOR.SYS fields
using the format's required 24-hour representation. First-time callers now
leave the previous-call field blank instead of repeating the current time.

## Changes

- Preserve the prior login timestamp before updating `LastLoginTs`.
- Render the two DOOR.SYS call-time fields independently in 24-hour format.
- Add focused regression coverage and a `WHATSNEW.md` entry.

## Testing

- `npx mocha test/record_login.test.js test/dropfile_comm_type.test.js --timeout 5000 --require test/setup.js --exit`
- `npm run lint`
- `npm run pretty:check`

`npm test` reached 2,453 passing tests but exited with 122 unrelated failures
in the BinkP/socket, shared-config, case-sensitive filesystem, and TTY-dependent
suites. The focused suites above pass independently.

### DIFF
--- a/WHATSNEW.md
+++ b/WHATSNEW.md
@@ -3,6 +3,8 @@ This document attempts to track **major** changes and additions in ENiGMA½. For
 
 ## 0.5.1-beta
 
+* **DOOR.SYS reports distinct call times** ([#824](https://github.com/NuSkooler/enigma-bbs/issues/824)) — the current and previous login times now occupy their respective fields in 24-hour format. First-time callers leave the previous-call field blank instead of repeating the current time.
+
 * **Post from a message list** ([#213](https://github.com/NuSkooler/enigma-bbs/issues/213)) — <kbd>P</kbd> on the message list, or the personal message list, starts a new message where before you had to back out to the message menu to reach its own <kbd>P</kbd>. The post goes to the area the highlighted message is in, so it does the right thing on a list that spans areas. An existing board keeps the menus it has, so see [UPGRADE.md](UPGRADE.md) for the binding to add.
 
 * **Doors can be handed a [`BBSDEV.DRP`](https://github.com/RealDeuce/bbsdev.drp) drop file** — set `dropFileType: BBSDEV` on an `abracadabra` door. The older formats are CP437, so a user whose name ENiGMA½ stores fine reaches the door mangled, and they leave the terminal's character set, the caller's language and the meaning of the comm field to convention. The new format is UTF-8 and states all three.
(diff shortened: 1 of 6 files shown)


NuSkooler (approved):
Thank you for your contribution @sjh9714, this is great!

🍻

Issue #831, opened 2026-09-13
On Enigma HQ, Fediverse search, cursor appears in the wrong place


andy5995:
<img width="1280" height="960" alt="Image" src="https://github.com/user-attachments/assets/4374a159-59ea-4246-a833-61e88db72905" />

Maybe related to https://github.com/NuSkooler/enigma-bbs/pull/790

I'm using synterm with ssh, on Manjaro Linux.

NuSkooler:
Should be resolved

PR #853, opened 2026-09-14
fix(activitypub): size the search results list to the space it has (#831)

NuSkooler:
The layout half of #831. No files in common with #852, so the two can land in either order.

## `VM3` is 7 rows too tall

`VM3` is configured `height: 14` in `theme.hjson`. The art places it at row 15 and the `TL10` status line at row 22, so the list runs rows **15–28**: result number 8 lands on top of the status line, and rows 26–28 clamp onto row 25 and blank out the bottom border of a 25-row art.

The space actually available is rows 15–21, so the height is **7**.

## Why `setHeight()` is not touched here

The underlying enabler is an asymmetry in `View`: `setWidth()` clamps to `termWidth - position.col`, but `setHeight()` clamps only to `termHeight` and ignores `position.row` — so nothing caught the overflow.

Deliberately left for its own change, for three reasons:

1. **It would not have fixed this.** A row-aware clamp at row 15 on a 25-row terminal gives **11**, which still overwrites `TL10` at row 22. The constraint here is the status line, not the terminal bottom.
2. **`setWidth()` — the model to copy — is itself off by one.** A view at column 6 gets 74 columns where cols 6–80 is 75. Copying its shape propagates that.
3. **`setPosition()` does not re-clamp, and `applyViewOffsets()` moves views after construction**, so any position-dependent height goes stale the moment a view is offset. That is a design question about where the clamp belongs.

All three verified by executing the real classes.

## A dead form entry on the actor view

Form 1 configured `MT7`, but the art places `%MT1` and `ap_search.js` reads the summary as view id `1` — three sources against one.

Because `MT7` does not exist, the entry was silently dropped and took `mode: preview`, `acceptsFocus: false`, `acceptsInput: false` and `hyperlinks: true` with it. The summary has therefore been a default `MultiLineEditTextView`: focusable, **accepting input**, in edit rather than preview mode, with no hyperlink rendering. This is the silent-shadow class of bug from the #281 / #783 work.

The form has no focusable view after this, which is fine — `_displayActorView()` calls `setFocus(true)` on the view controller, and that attaches client events regardless, so the escape / q / space / arrow action keys keep working.

## Tests

3 new tests in `test/activitypub_search_layout.test.js`, deriving MCI rows from the shipped art rather than hard-coding them. All three fail against the previous values:

- `VM3 height 14 from row 15 reaches row 28, overwriting the TL10 status line at row 22 (max height is 7)`
- `VM3 reaches row 28 but the art is only 25 rows`
- `form 1 configures MT7, which the shipped art does not place (art has: MT1, TL10, TL11, TL12, TL13, TL14, TL15, TL16)`

`2868 passing` vs `2865` on master, 9 pending, live suite 8 passing. Lint and `pretty:check` clean.

## Note

No art files touched. `config/menus/` is gitignored, so nothing was regenerated there.


NuSkooler:
Follow-up from a deeper look at the `setHeight()` question this PR defers. One correction to the PR body, one caveat on a test in this PR, and one new defect that is **not** fixed here.

## Correction: `setWidth()` is not off by one

The PR body says *"`setWidth()` is itself off by one (a view at column 6 gets 74 columns where 6-80 is 75)."* **That's wrong — the `-1` is deliberate wrap protection.**

`min(w, termWidth - col)` always yields exactly "spans `col`..79" on an 80-column terminal, at every column:

```
col  1 -> width 79 -> spans 1..79      col  6 -> width 74 -> spans 6..79
col  5 -> width 75 -> spans 5..79      col 56 -> width 24 -> spans 56..79
```

It reserves column 80. Three pieces of evidence that this is intentional:

- **No shipped art ever reaches column 80.** Max visible width across all 121 files in `luciano_blocktronics` is 79; 41 of them stop at exactly 79.
- **`disableVT100LineWrapping()` exists but has zero callers**, and its own comment warns clients do not reliably honour it. Auto-wrap is live.
- Writing the 80th cell leaves a pending wrap on most terminals — for a view that pads to its full width, that risks a scroll.

So `setWidth()` should be left alone and documented, not "fixed". Only the `setHeight()` side is genuinely missing a position-aware clamp, and there is already an in-tree precedent for it in `VerticalMenuView.autoAdjustHeightIfEnabled()` / `FullMenuView.autoAdjustHeightIfEnabled()`, both of which do `min(h, termHeight - position.row)`.

## Caveat on one of this PR's tests

`keeps the results list inside the art` encodes a rule that is **not generally valid**. Six shipped views legitimately extend past their art, because a short header art plus a list that fills the remaining screen is a normal pattern — e.g. `FDETLST.ANS` is 4 lines and its `VM1` is 17 rows.

It is correct for `activitypub_search_main.ans` specifically, which is a full-screen frame with a bottom border. Worth a comment so nobody promotes it to a general check later. The first test (`clear of the status line below it`) is the one carrying the real meaning.

## The deferral reasoning holds, and gets stronger

Two measurements support leaving `setHeight()` out of this PR:

- Across all 8 menu templates plus the theme — 298 configured dimensions, 260 resolved to real MCI positions using ENiGMA's own `ANSIEscapeParser` — **zero** heights exceed `termHeight - row` on 80x25. A clamp would be a no-op on everything currently shipped.
- Even pre-fix it would not have fixed this bug: `VM3` h=14 at row 15 clamps to 10 (current shape) or 11 (inclusive), where the correct value is 7. And on an 80x50 terminal the clamp does nothing at all while the overlap persists, since the art is 25 rows either way.

The binding constraint is the neighbouring view, not the terminal. A rectangle-intersection check over the shipped config finds **0 false positives** across 393 pairs, and catches this bug exactly when `VM3` is restored to 14 (`VM3 r15,c6 74x14` vs `TL10 r22,c6 65x1`). A row-only rule produces 48 hits, nearly all legitimate side-by-side layouts. That suggests overlap validation belongs with the menu/theme validation work rather than in `View`.

## New, unfixed: `VM3` is also too wide

Not addressed in this PR.

The theme configures `VM3` `width: 79` at column 6, so it is silently clamped to **74** today. Meanwhile `formatPeopleLine()` builds 78-character lines and `formatPostLine()` 79 — so result rows are being truncated by about 5 columns on this screen right now, independent of the height problem.

Twelve other views are similarly truncated, including `activityPubMsgViewer/1 MT1` (78 at col 5 -> 75) and `fileBaseListEntries/3 MT1` (79 at col 3 -> 77). Each needs a judgement on whether the theme value or the art position is the wrong one, so it is left for its own change.

PR #856, opened 2026-09-15
fix(offline mail): take the uploaded packet back instead of sending i…

andy5995:

Offline mail did not work on a stock install. Four separate defects, each hidden behind the one in front of it, all found by building a board with `oputil config new` and pressing <kbd>B</kbd>.

### The defects

**`fix(offline mail): take the uploaded packet back instead of sending it to the file base`**

`file_transfer_protocol_select` routes a completed recv to `fileBaseUploadFiles` unless the caller sets `extraArgs.returnToCaller` — `fse.js` does, and the import did not. So an uploaded reply packet went to the file base upload module, which found no upload area, logged that it had been re-entered without one, and returned having done nothing. The caller was left on `PROCESSING UPLOAD(S)` with no import and no way forward.

Setting that flag alone was not enough: the module also marked itself `MenuFlags.NoHistory`, and `MenuStack.goto()` pops a `NoHistory` module as it leaves, so it was off the stack by the time the transfer finished and `prevMenu()` landed a menu too far down. `upload.js` carries `NoHistory` safely because protocol selection re-enters it *by name*; anything relying on the return trip cannot. The flag is gone, with a comment saying why.

**`fix(offline mail): run on a menu that has no art`**

`misc/menu_templates` ships `bluewaveExport` and `offlineMailImport` with no art, and the docs show them that way, so this is the default path. With no art there is no MCI map, and `prepViewController` refuses one. Three layers:

* preparing views at all, which neither module needs without art;
* the handler for *that* error calling `temptmp.cleanup()` on a session a later step creates — turning a recoverable error into an uncaught `TypeError` that wedged the session;
* `_performExport` reaching through `this.viewControllers.main`, which that step never created. The code already guarded each individual view being absent, just not the controller.

`MenuModule.getCustomViewsWithFilter()` had the same assumption and is shared by every module that draws custom views, so it is fixed where it lives — `getCustomView()` directly above it already checks.

**`fix(offline mail): tell the caller what happened`**

Both modules reported through a status view that the shipped menus do not have, so an export or import ran to completion and returned to the message menu with nothing on screen — indistinguishable from the key doing nothing. They now write the outcome to the terminal and pause on it when there is no view. The export also returned on error without logging anything, so a genuine failure and "no messages to export with current criteria" looked identical from both the screen and the log.

**`fix(bluewave): report why the archiver failed`**

`_producePacketArchive` ignored `compressTo()`'s error and then `stat()`'d the archive it had not written, so a board with no `zip` installed was told `ENOENT ... ENIGMA.TU0` — a missing file rather than the missing archiver behind it.

### How it was verified

A board built from the templates, over Telnet from SyncTERM 1.9, running the full loop: export a packet, download it, read it and reply in NoCarrierMail 0.55, upload the reply.

```
02:37:47  offlineMailImport                        <- U
02:37:48  rz ... tempDir=/tmp/enigofflineimport-.../
02:37:56  Process exited exitCode=0
02:37:56  Protocol select received recv file paths: enigma.new
02:37:56  Goto menu module :: menuName=offlineMailImport     <- returns to the importer
02:37:56  Offline mail import complete :: imported=1, rejected=0
```

and the message it wrote:

```
message_id 2 | general | To: andy5995 | From: andy5995 | "Re: hello test"
reply_to_message_id = 1
meta: bw_echotag=GENERAL, bw_reply_to_num=1, local_from_user_id=1
```

`reply_to_message_id` is the threading from #822 resolving against a real message base for the first time.

### Notes

* Three of the four predate #822 and came in with #798/#799; only the archiver one is from the Blue Wave writer itself.
* This sits on top of #854, which fixed a different bug in the same two modules — `autoNext` racing their own `gotoMenu` — and was not this.
* The `returnToCaller` fix has a regression test. The other three do not: each needs a module's waterfall to run against config, art and file-area state, and a test that stubs all of that would assert its own stubs. A live board is what found them and a live board is what proves them, which is worth saying plainly rather than padding the diff with tests that cannot fail.


core/message_base_offline_import.js
```diff
@@ -175,7 +175,16 @@ exports.getModule = class MessageBaseOfflineImport extends MenuModule {
...
-        this.setMergedFlag(MenuFlags.NoHistory);
+        //
+        //  Deliberately NOT MenuFlags.NoHistory, which upload.js carries.
+        //  MenuStack.goto() pops a NoHistory module as it leaves, so this one
+        //  would be off the stack by the time the transfer finished, and the
+        //  prevMenu() that hands the upload back would land on the menu below
+        //  it instead -- no import, and the caller dropped a menu too far.
+        //  upload.js gets away with it because protocol selection re-enters it
+        //  by name rather than by returning. This module removes itself when
+        //  it is done, so it leaves nothing behind either way.
```
NuSkooler, line 186: **Blocking:** it does leave something behind — itself, while protocol selection is running — and that turns ESC into a loop with no way out.

Keeping this module on the stack is what makes the happy path work, but `fileTransferProtocolSelection` binds ESC to `@systemMethod:prevMenu` (`misc/menu_templates/file_base.in.hjson`, and the dev config menu does the same). `MenuStack.prev()` pops protocol selection and re-enters *this* module with `lastMenuResult = protocolSelect.getMenuResult()`, which is `undefined` when nothing was received. So `isFileTransferComplete()` is false and `finishedLoading()` calls `_receivePacket()` — straight back to protocol selection.

Stubbing it out at the prototype level:

```
restored tempRecvDirectory: /tmp/enig-round-1/
isFileTransferComplete(): false
finishedLoading() on that re-entry does:
  -> gotoMenu(fileTransferProtocolSelection): protocol selection, again
```

Before this PR the `NoHistory` pop meant ESC landed back on the message menu. Now the only exit is to pick a protocol and abort the transfer — that yields `recvFilePaths: []`, which is truthy and completes the round trip. Every cycle also `mkdir`s a fresh temp dir and abandons the previous one.

The discriminator is already here: `restoreSavedState()` puts `tempRecvDirectory` back on re-entry, and it is only ever set once `_receivePacket` has run. So in `finishedLoading()`, before the `_receivePacket` call:

```js
if (this.tempRecvDirectory) {
    //  re-entered from protocol selection with nothing: the caller backed out
    return this._finish();
}
```

That is testable exactly the way the new `returnToCaller` test is.

core/message_base_offline_export.js
```diff
@@ -123,6 +123,17 @@ module.exports = class MessageBaseOfflineExport extends MenuModule {
...
                 [
                     callback => {
+                        //
+                        //  A menu with no art produces no MCI map, and a view
+                        //  controller refuses to load without one. The menu
+                        //  this module ships on carries no art, so preparing
+                        //  views unconditionally failed the export before it
+                        //  started.
+                        //
+                        if (!mciData.menu) {
```
NuSkooler, line 137: This guard only gets Blue Wave through. The next waterfall step still runs `validateMCIByViewIds('main', required)`, which returns `Form does not exist: main` when no controller was created — so skipping `prepViewController` just moves the failure down one step.

Blue Wave escapes because it overrides `requiredViewIds()` to `[]`. QWK inherits the base `[status, progressBar]`:

```
Blue Wave export (overrides requiredViewIds -> []): runs without views
QWK export (inherits the base requiredViewIds): FAILS -- Form does not exist: main
```

So a QWK export on an art-less menu still bails before it starts, now silently apart from the new log line below.

Fold the validation into the same guard — `if (!mciData.menu) return callback(null)` covering both steps — which is what `message_base_offline_import.js` already does in its own `mciReady()`.

core/message_base_offline_export.js
```diff
@@ -169,9 +180,23 @@ module.exports = class MessageBaseOfflineExport extends MenuModule {
...
+                    }
 
                     if (err) {
+                        //  said out loud: this used to return the caller to
+                        //  the previous menu with no record of why, which
+                        //  looks from the outside like the export doing
+                        //  nothing at all
+                        this.client.log.warn(
+                            { error: err.message, reasonCode: err.reasonCode },
+                            `${this.packetFormatName} export failed`
```
NuSkooler: This line can throw, in the same shape as the `temptmp` bug fixed eleven lines above it.

`packetFormatName` is a throwing getter on the base class, and `_missingHook()` exists precisely because a format may not have supplied it. When it hasn't, `_performExport()` returns that error, we land here, and the template literal re-throws it — uncaught, out of the waterfall's final callback:

```
_missingHook() returned: Missing parameter(s): packetFormatName is required
THROWS inside the error handler: Missing parameter(s): packetFormatName is required
```

Only bites an incomplete third-party format, not the two shipped ones, but it turns a recoverable error into a wedged session — which is the thing this commit is fixing. `_performExport()` has already logged that case, so the simplest fix is to not reach through the getter here.

core/message_base_offline_export.js
```diff
@@ -189,7 +214,13 @@ module.exports = class MessageBaseOfflineExport extends MenuModule {
     }
 
     finishedLoading() {
-        this.prevMenu();
+        if (!this.finalStatus) {
+            return this.prevMenu();
```
NuSkooler: The one case the caller most needs told is the one that still says nothing.

Success and `NothingToDo` set `finalStatus` and get the terminal write plus the pause. A genuine error goes down the other branch in `mciReady()` — `prevMenu()` without ever calling `cb`, so `finishedLoading()` never runs at all — and the caller is dropped back on the message menu with a blank screen. That is the exact symptom the commit message describes, and only the log improved.

Worth routing the error branch through the same write-and-pause.

core/message_base_offline_import.js
```diff
@@ -629,6 +648,24 @@ exports.getModule = class MessageBaseOfflineImport extends MenuModule {
...
+        //  returned to the message menu with no word of whether anything was
+        //  posted. Say it on the terminal instead, and hold it there long
+        //  enough to read.
+        //
+        if (this._statusView()) {
+            return this.prevMenu();
+        }
+
+        const { imported, rejected } = this.summary;
+        const summary = rejected
```
NuSkooler: `finishedLoading()` routes a `_receivePacket` error into `_finish()` too, so a failed `temptmp.mkdir()` tells the caller `Imported 0 message(s)` — the import found nothing, rather than the upload never having started. Worth a separate line for that path.

core/bluewave_mail_packet.js
```diff
@@ -711,7 +711,15 @@ class BlueWavePacketWriter extends EventEmitter {
                     packetPath,
                     files,
                     this.workDir,
-                    () => {
+                    err => {
+                        //  Report why the archiver failed. Ignoring it left
+                        //  the stat below to report ENOENT on the archive
+                        //  that was never written, which says nothing about
+                        //  a missing or misconfigured archiver.
+                        if (err) {
```
NuSkooler, line 719: `qwk_mail_packet.js` has this verbatim — same ignore-the-error-then-`stat()` pattern, same `ENOENT` on an archive that was never written. Worth taking both while you are in here; it is the same six lines.

NuSkooler (commented):
Went through this one carefully. The four diagnoses are all correct — I confirmed `MenuStack.goto()` really does pop a `NoHistory` entry, that `this.temptmp` was genuinely undefined for any error before waterfall step 3, and that `compressTo()` does report `Unknown archive type` and spawn failures. The `getCustomViewsWithFilter()` guard is necessary rather than just tidy: the export's progress handler calls `updateCustomViewTextsWithFilter('main', ...)` and on the new no-art path `viewControllers.main` is genuinely undefined. Test suite is 2889 passing / 0 failing, eslint and prettier clean.

One blocker and two things I'd like fixed before this lands, plus three smaller notes inline.

On "a live board is what found them and a live board is what proves them" — fair for the export waterfall, but the first two findings below both fall out of plain prototype calls, no config or art or file-area state involved. Worth another look at what is actually testable here.

NuSkooler (commented):
All six land, and the two you found on your own are good ones — `Errors.UserInterrupt` in particular, since ESC during an export was previously indistinguishable from a crash. `optionalMoveToPosition()` reaches well past this PR (`abracadabra.js` x4, `file_transfer.js:662`, `upload.js:674`, `gotoMenuOrShowMessage`), but every one of those writes a message or paints art immediately before pausing, so drawing at the cursor is right in all of them, and the seven new tests cover it. 2899 passing, eslint and prettier clean.

The download-queue key is a good addition — keep it. The `MSGMNU.ANS` edit holds the original 31-column offset exactly (`\x1b[5C` + the label + `\x1b[10C`) and matches the bright-key/cyan-label scheme of every other entry, so no rework needed there.

One thing left, plus three small ones inline.

core/message_base_offline_export.js
```diff
@@ -335,16 +382,31 @@ module.exports = class MessageBaseOfflineExport extends MenuModule {
...
-        );
+        //
+        //  What the caller is told at the end. With a status view they have
+        //  been watching it all along; without one -- which is what the
+        //  shipped menu is -- the export would otherwise finish in silence.
+        //
+        this.finalStatus = null;
+        const finalStatus = status => {
+            updateStatus(status);
+            if (!statusView) {
```
NuSkooler: This guard is now what stops `finishedLoading()` from doing what its new comment says.

Line 231 promises to hold the outcome "either way: in the status view on a menu with art, on the terminal on one without" — but `finalStatus` still declines to set `this.finalStatus` when a status view exists, so on a menu with art `finishedLoading()` sees it null and returns `prevMenu()` straight away. Replaying the helper and `finishedLoading()` as they are:

```
no art     : status view text = null | term: A Blue Wave packet has been placed in your download queue  pausePrompt() -- held for the caller
with art   : status view text = "A Blue Wave packet has been placed in your download queue" | prevMenu() immediately -- nothing held
```

So success, "No messages to export with current criteria" and the new "Export canceled" all still flash past with art. The only thing that reaches the new `statusView.setText()` branch is a failure, because line 218 sets `finalStatus` unconditionally.

Dropping `if (!statusView)` here makes it true in both cases — `updateStatus()` on the line above has already written the text, so the `setText()` at line 235 is just a harmless re-set.

core/message_base_offline_export.js
```diff
@@ -180,16 +213,30 @@ module.exports = class MessageBaseOfflineExport extends MenuModule {
                             );
                         }
 
-                        return this.prevMenu();
+                        //  a failed export used to return the caller with no
+                        //  word of why; finishedLoading() says it
+                        this.finalStatus = 'The export failed -- see the log';
                     }
-                    return cb(err);
+                    return cb(null);
```
NuSkooler, line 227: Two small things fall out of always calling back here, and a third nearby:

- `displayPauseIfRequested` now runs on every path, so a sysop who set `pause: true` on the export menu gets that pause *and* the one `finishedLoading()` adds. Not the shipped config, but it used to be one.
- The NORESULTS branch just above is now the only path that never calls `cb` at all. It is still dead, so nothing breaks today, but if that `:TODO:` is ever picked up it will stall the init sequence — worth a `return cb(null)` after the `gotoMenu()` while the reasoning is fresh.
- Unrelated leftover: `const mainVc` on line 387 now only serves the progress bar, since the refactor moved status views to `this.getView()`. `this.getView('main', MciViewIds.main.progressBar)` would let it go.

NuSkooler: NON-BLOCKING

core/message_base_offline_export.js
NuSkooler: Only blocker left is here I believe.

NuSkooler (approved):

Issue #719, opened 2026-08-19
Writer/reader disagree on which network's outbound dir gets the "default" (unsuffixed) outbound/ folder


jriethoven:
Files involved: core/scanner_tossers/ftn_bso.js (writer) and core/binkp/bso_spool.js (reader, used by core/binkp/caller.js)

The bug:

Both files determine which configured FTN network is "default" (and therefore gets its outbound mail written to/read from the bare mail/ftn_out/outbound/ directory, vs. a mail/ftn_out/<networkName>/ subdirectory) — but they use two different, disconnected implementations that disagree whenever more than one network is configured.

Writer (ftn_bso.js, getDefaultNetworkName()):

js
if (this.moduleConfig.defaultNetwork) {
    return this.moduleConfig.defaultNetwork;
}
const networkNames = Object.keys(config.messageNetworks.ftn.networks);
if (1 === networkNames.length) {
    return networkNames[0];
}
// otherwise falls through and returns undefined

With 2+ networks configured and no explicit defaultNetwork set, this returns undefined. Every network is then treated as non-default and written to mail/ftn_out/<networkName>/.

Reader (bso_spool.js, _defaultNetworkName()):

js
_defaultNetworkName() {
    const names = Object.keys(this._networks);
    return names.length > 0 ? names[0] : null;
}

This always returns whichever network is listed first in messageNetworks.ftn.networks, regardless of how many networks exist and regardless of any defaultNetwork config. It never checks moduleConfig.defaultNetwork at all.

Effect: With 3 networks configured (fidonet, fsxnet, spooknet — in that config order) and no defaultNetwork set:

Writer puts fidonet mail in mail/ftn_out/fidonet/ (correctly, per its own logic — no default exists).
Reader (getNodesWithPendingMail() / _allOutboundDirs()) looks for fidonet mail in mail/ftn_out/outbound/ (since fidonet is first-listed), never finds it.

Result: outbound netmail/echomail for the first-listed network is silently invisible to any pending-mail scan (BinkP Poll, scheduled polls via caller.js), and just sits in the spool indefinitely with no error logged.

Workaround found: explicitly setting scannerTossers.ftn_bso.defaultNetwork: <name> fixes the writer to agree with whatever the reader already assumes (first-listed network) — but only works as long as that network stays first in messageNetworks.ftn.networks. It's fragile: reordering/adding networks above it silently reintroduces the mismatch, now against whichever network becomes newly-first.

[...]

PR #863, opened 2026-09-15
fix(bso): prune a forwarded payload and the TIC announcing it together

### LINKED ISSUE #862: BSO: `oputil bso prune` leaves a queued TIC announcing the payload it just removed
Split out from the queue-hygiene discussion under #743 and #755, but this one is a defect in shipped behaviour rather than a feature.

## What happens

`oputil bso prune <address> --yes` removes a flow file reference whose payload has gone missing, but leaves the generated `.tic` that announces it queued — and the TIC file itself on disk.

`pruneMissingRefs()` only removes entries whose *file is gone*. When a forwarded payload vanishes from the file base, that entry resolves to nothing and is correctly reported `missing`. The `.tic` we generated for it is still sitting in the outbound, so it resolves fine and stays `pending`. Pruning takes the payload and leaves the announcement.

The downlink then collects a TIC describing a file that is never sent. That is the #735 situation — an orphaned control file with no payload — except we have inflicted it on someone else's system, under our `From`, our `Path` and a `Seenby` containing us.

## Why

Two implementations of "remove a reference from a flow file" disagree about whether the companion TIC exists.

`scrubFlowFileRefs()` (`core/scanner_tossers/ftn_bso.js`) — the `Replaces` dequeue path — knows the rule and states the invariant in its own comment:

> We queue a payload and its TIC in one append, adjacent and in that order, so a `^...tic` line immediately following a matched payload belongs to it and goes too -- otherwise it would sit in the outbound forever with nothing referencing it.

It removes the line and unlinks the file.

`_pruneFromFlowFile()` (`core/binkp/bso_spool.js`) — what `oputil bso prune` calls — splices matched lines by index and rewrites the file. It has no notion of a companion at all, and `core/oputil/oputil_bso.js` contains no reference to `.tic`.

Both are correct in isolation; the rule simply lives in only one of them.

## Repro

1. A `ticAreas` entry with a downlink that is not polling.
2. Import a TIC for that area. The downlink's flow file gets `<payload>` then `^<generated>.tic`, adjacent.
3. Remove the payload from the file base.
4. `oputil bso status` reports 1 missing for that node; `oputil bso prune <addr> --yes` removes it.
5. The `^...tic` line is still in the flow file and the `.tic` is still on disk. The next session ships the TIC alone.

## Fix

Teach the prune path the adjacency rule. Since the point of the exercise is that the two must not disagree, the rule belongs in one shared place (`core/bso_util.js` already exists so the flow file's writer and reader cannot drift) rather than being written a second time.

NuSkooler:
Fixes #862.

## The defect

`oputil bso prune <address> --yes` removed a flow file reference whose payload had gone missing and left the generated `.tic` announcing it queued — and the `.tic` itself on disk.

`pruneMissingRefs()` removes entries whose *file is gone*. Only the payload can go missing; the generated TIC sits in the outbound where we wrote it, so it resolves fine and stays `pending`. Pruning reliably took one and left the other.

The downlink then collects a TIC describing a file that is never sent. That is #735 — an orphaned control file with no payload — except inflicted on someone else's system, under our `From`, our `Path` and a `Seenby` containing us.

## Why it happened

Two implementations of "remove a reference from a flow file" disagreed about whether a companion exists.

`scrubFlowFileRefs()` (the `Replaces` dequeue path) knew the rule and stated the invariant in its own comment — *"We queue a payload and its TIC in one append, adjacent and in that order"* — removed the line and unlinked the file.

`_pruneFromFlowFile()` spliced matched lines by index and rewrote the file, with no notion of a companion. `oputil_bso.js` contained no reference to `.tic` at all.

Both were correct in isolation. The rule simply lived in only one of them.

## The fix

The rule now lives once, in `core/bso_util.js` — which exists precisely so the flow file's writer and reader cannot drift (#719) — as `withCompanionTicRefs()`, plus the `flowRefIsSent` / `flowRefBody` / `companionTicIndex` primitives underneath it. Both paths call it; `scrubFlowFileRefs` is refactored onto it rather than keeping its own copy.

Prune additionally:

- **unlinks the dequeued TIC.** `^` is delete-after-send, and with its reference gone nothing will ever send it — it would sit in the outbound forever. Best effort, so a failed unlink does not cost the rewrite, which is the half that actually stops the downlink receiving it.
- **reports the companion distinctly** (`   (TIC announcing it)`). An operator who asked to drop one missing file and sees two paths needs to know the second was not a second missing file.

Unchanged and deliberately so: a `~` (already sent) TIC is never touched — it has gone, and unlinking the file destroys the record. The dry run still reports only what `inspectOutbound()` found, since it never reaches the rewrite; the help text and docs now say more lines may go than it listed.

## Tests

+21 (2887 → 2908), 0 failing.

- `test/bso_util.test.js` — the rule itself: adjacency, the `^` directive requirement, case-insensitive extension, a sent companion, a payload that is last, `noTic` links with no companion at all, and several forwards in one flow file pairing correctly.
- `test/binkp_bso_spool.test.js` — the defect end to end at the spool: dequeued, unlinked, reported, and the ordinary case (payload still present) left byte-for-byte alone.
- `test/oputil_bso.test.js` — through the real CLI process, since the whole defect was that this path and `ftn_bso`'s disagreed.

Each was mutation-checked: disabling the companion expansion fails 9 tests **including the pre-existing `scrubFlowFileRefs` case**, which confirms the shared helper is genuinely load-bearing for both callers rather than a shim. Removing the unlink fails 3; removing the companion reporting fails 2.


NuSkooler:
## Review fixes

Five defects found on review, closed in `15f452d6`. The first defeated this PR's own fix in exactly the case the surrounding code goes out of its way to handle.

**1. The TIC was unlinked by the raw text of its reference.** `resolveFlowRef()` deliberately falls back to the referenced *basename* beside the flow file when the stored absolute path no longer resolves — a relocated outbound, or a sysop recovering mail filed in the wrong directory. So when that fallback was what found the file, the line was dequeued and the `.tic` was left on disk: precisely the orphan this PR removes. The operator was also shown a path that did not exist, while every sibling row showed the resolved one.

**2. Nothing checked where the unlink pointed.** A generated TIC is always written to the same directory as the flow file that references it, so a sibling is both the correct fallback and a bound — `companionTicPath()` cannot return anything outside that directory. The reference is the one part of a flow file we did not necessarily write, and `oputil` runs as whatever user the operator is, from an arbitrary working directory.

**3. The unlink ran before the rewrite.** A rewrite that then failed (ENOSPC, EROFS, EACCES) left the flow file naming TICs that no longer exist — the dangling reference #735 is about — while the caller reported the entries as merely *busy*. Rewrite first: a failure now leaves everything as it was, and a failed unlink leaves litter rather than a broken queue.

**4.** The prune report appended companions after all payloads, so with two missing files the CLI's `(TIC announcing it)` named the wrong one. Now in line order.

**5.** `/\.tic$/i` treated a dotfile named `.tic` as having that extension, which the `extname()` test it replaced did not — a behaviour change in something described as a pure refactor.

### Two test gaps it exposed

`scrubFlowFileRefs` has two endings — unlink the flow file when nothing live is left, or rewrite it with the survivors. **Every existing case left nothing live**, so the rewrite branch could have been deleted outright without failing anything. And nothing covered a relocated outbound tree, which is finding 1.

### Verification

Full suite **2912 passing, 0 failing**. Five mutations, one per fix, each caught — including the rewrite-branch one, which survived until the new test was added.

Also worth recording what the review *cleared*, since it was the thing I was least sure of: it enumerated all three `flowFileAppendRefs` call sites and confirmed the adjacency invariant holds — only TIC forwarding emits `^…tic`, always as one append under the `.bsy`; `_applyFlowDispositionLocked` rewrites a line in place so indices are preserved; a `.tic` carried as a *payload* is queued with no directive and so is never claimed as a companion; and a `noTic` downlink has no `^…tic` lines at all.

Issue #753, opened 2026-09-02
TIC: passthrough file areas — carry an echo without storing it


NuSkooler:
Follow-on from #743. `ticAreas` requires a local file base area (`areaTag`), and the `store` stage hard-fails without one. So a hub can only carry an echo it also stores locally.

A hub carrying 40 echoes for its downlinks should not have to keep 40 local file areas, with the disk and file-base clutter that implies.

## Prior art

- htick: `PassFileAreaDir` plus areas declared passthrough — the file lands in a transit directory, is forwarded, and is removed.
- Mystic: tosses straight to downlink fileboxes.

## Scope

- A `ticAreas.<tag>.passthrough: true` (or an absent `areaTag` meaning the same), with a spool directory for transit files.
- Forwarding reads from that directory instead of file base storage. #743 already forwards the *stored* copy rather than the inbound one, so this is mostly a matter of where "stored" points.
- Cleanup once every downlink has collected it — which interacts with queue hygiene (see the related issue): a transit file must outlive the last pending send.
- `Replaces` still needs to dequeue superseded transit files.

Note that the dupe protection #743 relies on is the file-base collision rename, so a passthrough area needs its own answer for a re-announced file.

Issue #743, opened 2026-08-29
TIC-announced files are never forwarded to downlinks (file echo forwarding / re-hatching unimplemented)


NuSkooler:
Component: TIC processing in `ftn_bso.js` / `tic_file_info.js`

## Summary

ENiGMA½ can only ever be a **leaf node** for FTN file echoes. A TIC-announced file received from an uplink is imported into the local file base and then deleted from the inbound; it is never passed on to any downlink. There is no configuration to declare downlinks for a TIC area, no TIC file is ever generated, and the `Path` / `Seenby` fields are parsed but never written.

This is flagged in the source, at `core/scanner_tossers/ftn_bso.js:2877` (`processSingleTicFile`, immediately before the `cleanupOldFile` waterfall stage):

```js
//  :TODO: from here, we need to re-toss files if needed, before they are removed
function cleanupOldFile(localInfo, callback) {
```

The comment's "before they are removed" is the load-bearing part: by the time that stage runs, the file has been copied into file base storage, and `removeAssocTicFiles()` unlinks the inbound `.tic` and its payload as soon as `processSingleTicFile` returns. Any forwarding has to happen while both are still on disk.

## Confirmed missing

Verified across the tree — none of this exists today:

* **No TIC generation.** Nothing anywhere writes a `.tic` file. `TicFileInfo` is read-only: `createFromFile()` parses, and there is no serializer.
* **No `Path` / `Seenby` maintenance.** Both are parsed in `TicFileInfo.createFromFile()` (`case 'seenby':`) and are otherwise unused. FTS-5006 requires a forwarding processor to append its own address to `Path` and add itself plus its downlinks to `Seenby`; without that, forwarded files would loop.
* **No downlink configuration.** `scannerTossers.ftn_bso.ticAreas` maps an external area tag to a *local* file base area (`areaTag`, `storageTag`, `hashTags`) and nothing else. There is nowhere to say "and send this area on to 21:1/200".
* **No outbound queueing for files.** The BSO/BinkP outbound path queues mail (packets and bundles); nothing queues an arbitrary file plus a companion `.tic` to a node.

The documentation does not claim otherwise — `docs/_docs/filebase/tic-support.md` describes inbound import only — so this is a missing feature rather than a broken one.

## Impact

[...]

PR #865, opened 2026-09-15
feat(tic): hatching — originate a file into a file echo we carry

### LINKED ISSUE #751: TIC: hatching — originate a file into a file echo
Follow-on from #743, which added forwarding. ENiGMA½ can now pass a file echo *on*, but still cannot **originate** one: there is no `hatch` anywhere in the tree. A hub that cannot hatch cannot ship its own nodelist, infopack or announcements — arguably the point of running one.

## Scope

Most of the work already exists. #743 added `core/tic_file_writer.js` and the per-downlink queueing (`forwardTicToDownlinks` → `forwardTicToOneDownlink`), deliberately structured so hatch is a thin caller: build a TIC from local metadata instead of from an inbound one, then reuse the same writer, the same unique-8.3 naming, and the same `flowFileAppendRefs` payload-then-TIC append.

**The part that is not shared is the operator surface, and that is the real design question this issue should settle.** Options, not mutually exclusive:

- `oputil.js fb hatch <areaTag> <file> [--desc ...] [--replaces ...]` — at minimum. Fits existing oputil file-base commands and is scriptable, which is what a nodelist cron job wants.
- A menu/ACS-gated action so a sysop can hatch from the board.
- A watched directory per area, so dropping a file in hatches it.

Whoever picks this up should decide that first; the TIC generation itself is largely done.

## Notes

- `Origin` is us, `Path` starts with our line, `Seenby` is us plus every downlink.
- `Replaces` on hatch should dequeue the superseded file from downlinks that have not collected it — the machinery for that already exists (`dequeueReplacedForDownlinks`).
- htick's equivalent is its `hatch` command reading a `.tic`; Mystic hatches via MUTIL.

### LINKED ISSUE #864: TIC: a wildcard `Replaces` never matches, so allowReplace silently duplicates instead of superseding
Found while building hatching (#751), which needs the same lookup. Verified against `master` @ 9e0a49a9.

## What happens

A TIC carrying a wildcard `Replaces` never matches the file it supersedes. With `tic.allowReplace` on, the import silently stores a **second** entry alongside the old one instead of replacing it, and — because `dequeueReplacedForDownlinks()` is only reached when an existing entry was found — never dequeues the superseded file from downlinks that have not collected it.

So a weekly nodelist accumulates: `NODELIST.239`, `NODELIST.246`, `NODELIST.253`, all live in the file base, and a downlink that polls fortnightly receives every one of them. That is precisely the outcome `scrubFlowFileRefs()` was written to prevent, and it has never been reachable through a wildcard.

An exact `Replaces NODELIST.246` works. `Replaces NODELIST.*` — the form the field exists for, and the form every nodelist hatch in the field uses — does not.

## Why

`FileEntry.findFiles()` (`core/file_entry.js`, the `metaPairs` / `wildcards` branch) converts the glob and *then* sanitizes it:

```js
mp.value = mp.value.replace(/\*/g, '%').replace(/\?/g, '_');
appendWhereClause(
    `... meta_value LIKE '${sanitizeString(mp.value)}'`
);
```

`sanitizeString()` (`core/database.js`) escapes `%` as `\%`, because it is built for *literal* values where a stray `%` would otherwise act as a wildcard. Applied after the conversion it escapes the wildcards the conversion just introduced:

```
NODELIST.*   ->  NODELIST.%   ->  LIKE 'NODELIST.\%'
```

SQLite's `LIKE` has no default escape character — `ESCAPE` must be given explicitly — so that backslash is a literal backslash. The query asks for a file whose `short_file_name` is exactly `NODELIST.\%`, and matches nothing.

`?` survives, since `sanitizeString()` does not escape `_`. So `NODELIST.2?6` works and `NODELIST.*` does not, which is a good way to confirm the diagnosis.

Two escaping layers meet here and only one was applied, in the wrong order.

## Also

`mp.value = ...` mutates the caller's `metaPairs` object in place, so a filter reused across two calls is converted twice. Harmless today (no caller reuses one) but it is why the bug reads as a single-shot conversion.

## Fix

Escape the caller's literal `%`, `_` and `\` *first*, then introduce the glob's own wildcards, and give the `LIKE` an explicit `ESCAPE` clause. Don't route a live pattern through `sanitizeString()`, which is for literals.

Worth noting separately: `sanitizeString()` escapes `\` and `%` in the MySQL style, and SQLite's only string-literal escape is a doubled `'`. That is not causing a problem elsewhere, but it means the helper cannot be the right tool for a pattern.

NuSkooler:
Closes #751. Also fixes #864, which this work uncovered — see below.

## What this adds

```bash
oputil.js fb hatch <areaTag> <file> [--desc ...] [--ldesc ...] [--replaces ...]
                                    [--tags ...] [--storage-tag ...] [--dry-run]
```

Forwarding passes on a file an uplink sent us; **hatching** puts one of our own into an echo. A hub that cannot hatch cannot ship its own nodelist or infopack — it can only repeat what it is told.

The area tag is the **FTN** one (a `ticAreas` key), not the local file base area tag: that is what an operator configuring a hatch is looking at, and one local area may carry several echoes.

The file is copied into the echo's local area, scanned and persisted, then announced to every downlink. `Origin` is us, `Path` begins with our line, `Seenby` is us plus every downlink.

## How it is built

Deliberately thin — none of #743's machinery is re-implemented.

`forwardTicToDownlinks()` is split so the half **after** its gates, `announceTicToDownlinks()`, is shared. What differs between a forward and a hatch is entirely upstream: a forward must pass `canForwardTic()` because a third party is asking us to re-announce their file under our name, while a hatch is the operator at the console and has no sender to authenticate. Everything below that line — the network and our address for it, the loop guard, the `Replaces` dequeue, the per-downlink TIC, the payload-then-TIC append — is identical and now literally the same code.

The TIC is **rendered as text and read back through `TicFileInfo`'s own parser** rather than assembled as an object (`createFromString`, split out of `createFromFile` in its own commit). That costs a parse we could have skipped and buys the property that matters: a hatched TIC reaches `TicFileWriter` in exactly the shape a forwarded one does — same `rawLines`, same conversions — so there is one way a TIC gets written and no second path to keep in step.

## `--replaces`, and the bug it uncovered

`--replaces NODELIST.*` is what makes a weekly nodelist cron work: it supersedes the previous file from the same origin in the same area, updates the entry in place, dequeues the old file *and its TIC* from downlinks that have not collected them, and removes the old physical file.

**It did not work, and neither did the same thing on the TIC import path.** `FileEntry.findFiles()` converted the glob's `*` to `%` and then ran the result through `sanitizeString()`, which escapes `%` because it is built for literal values:

```
NODELIST.*  ->  NODELIST.%  ->  LIKE 'NODELIST.\%'
```

SQLite's `LIKE` has no default escape character, so that backslash was literal and the query asked for a file named exactly `NODELIST.\%`. With `tic.allowReplace` on, a wildcard `Replaces` therefore stored a **duplicate** rather than superseding — and because the downlink dequeue only runs when an existing entry was found, the old file stayed live in the base and stayed queued for every downlink. A weekly nodelist accumulated, and a fortnightly downlink received every one of them.

An exact `Replaces NODELIST.246` worked, and so did `NODELIST.2?6` (`sanitizeString` does not escape `_`), which is what made it so quiet. Filed as #864 and fixed in the first commit here, separately reviewable — `--replaces` is untestable without it.

## Refusals rather than guesses

- An echo with no `downlinks` — a hatch would announce to nobody.
- An echo with no `areaTag` — passthrough areas are #753.
- A `Replaces` matching more than one file. Picking wrong deletes a real file locally *and* pulls it out of every downlink's queue.
- A name already in the area without `--replaces`. This one also **takes the renamed copy back out**, so a refused hatch leaves nothing in the area's storage directory. Announcing one name while shipping another is the orphan a collision-renamed import is already never forwarded for.

## Verified against a live instance

Not just tests. A throwaway ENiGMA with a real file base, `ticAreas` and outbound was driven through `oputil fb hatch` twice. What landed:

```
Created ENiGMA1/2 0.5.1.b (linux; x64; 22.22.2)
Area FSX_NODELIST
Origin 21:1/151
File NODELIST.246
Lfile nodelist.246
Crc 57F401D4
From 21:1/151
To 21:1/200
Pw DOWNPASS
Path 21:1/151 1789502393 Tue, 15 Sep 2026 19:59:53 GMT ENiGMA1/2 ...
Seenby 21:1/151
Seenby 21:1/200
```

…with the payload queued first and the TIC second under `^`, per FSC-0087. The second hatch with `--replaces NODELIST.*` left exactly one payload, one TIC and one database row.

Running the app also caught a miss the tests did not: `ticAreas.<tag>.areaDesc` was a new config key with no `core/config/meta.js` entry, so it reported `unknown key` at startup. Registered, schema regenerated, docs table updated.

## Tests

+49 (2887 → 2936), 0 failing.

- `test/tic_hatch.test.js` — the 8.3 name, the rendered TIC, and echo resolution.
- `test/ftn_tic_hatch.test.js` — what lands in the outbound: `Origin`, a `Path` with exactly our line, `Seenby` seeded with us and every downlink, per-downlink `From`/`To`/`Pw`, the loop guard still skipping our own address, and the supersede path dequeuing the old pair while leaving an already-sent reference alone.
- `test/file_base_db.test.js` — the #864 fix against a real SQLite file base.

Mutation-checked throughout. One case is worth calling out: the injection tests initially caught only 1 of 5 mutations, because my line-splitting helper split on CRLF and so could not see a **bare CR** — which is the dangerous one, since every C implementation splitting on `"\r\n"` treats it as a terminator. Rewritten to parse with the real reader and to assert no control character survives; all 7 now fail when the guard is removed.


NuSkooler:

| Pair | Result |
|---|---|
| #863 + #865 | clean |

Issue #759, opened 2026-09-02
TIC: find a payload a mailer renamed, by Size and Crc


NuSkooler:
Recorded from the #743 work as deliberately deferred. It looks like a small fix and is not one — the design questions below are the reason it was not simply done.

## What the spec asks for

FSC-0087, "Transmission of Files":

> File Forwaders should be transparent to the renaming of file by mailers. This means that if the mailer renames a duplicate file by renaming or bumpinmg a numeric extension, the File Forwarder should be able to use the size and crc fields of the TIC to find and properly rename the associated file referred to in the TIC.

*(typos are in the original)*

The scenario: a mailer receives `FOO.ZIP` into an inbound that already holds one, so it saves it as `FOO.ZI1`, `FOO.001` or similar. The TIC still announces `FOO.ZIP`, so the announced name is not on disk under that name.

## What we do today

`TicFileInfo.resolveFilePath()` (`core/tic_file_info.js`) tries the exact name, then falls back to a case-insensitive scan of the TIC's own directory — matching htick's `adaptcase()`. It does **not** look for a renamed file.

So such a TIC is held (per #735), retried each pass, and eventually given up on after `tic.holdMaxAgeMs` and rejected. The payload is left in the inbound.

## Why this is not a quick fix

**It sits on a hot path.** #735's whole design is that an unresolved TIC is *held* and retried on every import pass, for up to 48 hours by default. A name-miss fallback that scans the inbound and hashes candidates would therefore re-scan and re-hash on every pass, for every held TIC — and a large echo's inbound is not small. The naive implementation makes the common, correct case (payload genuinely has not arrived yet) expensive.

**It can match the wrong file.** Size plus CRC-32 is a weak identifier. Two announcements of similarly-sized files in a nodelist echo colliding on CRC-32 is not far-fetched, and mis-pairing a TIC with the wrong payload is worse than failing to pair it at all — the wrong file would be imported and, with #743, forwarded to downlinks under our name.

**There is no evidence it happens here.** A corpus of 1,769 real TICs from a live system showed payloads matching their announced names; spot checks against archived payloads confirmed CRC and size agreed. So this is spec conformance rather than a fix for something observed.

## Questions a design should answer

- When to attempt it: only on the *final* pass before `holdMaxAgeMs` expires, rather than every pass? That keeps the hot path cheap and only pays the cost when the alternative is rejection.
[...]

PR #866, opened 2026-09-15
feat(tic): passthrough file areas — carry an echo without storing it

### LINKED ISSUE #753: TIC: passthrough file areas — carry an echo without storing it

NuSkooler:
Closes #753.

## The blocker

A hub carrying forty file echoes for its downlinks had to keep forty local file areas, forty storage directories, and forty areas' worth of files its own users would never browse. `store()` called `getFileAreaByTag(localInfo.areaTag)` and hard-failed `Errors.UnexpectedState` without one.

```hjson
ticAreas: {
    fsx_gen: {
        passthrough: true
        network: fsxnet
        uplinks: [ "21:1/100" ]
        downlinks: [ "21:1/200", "21:1/300" ]
    }
}
```

An absent `areaTag` means the same thing — there is no local area to store into, which is what passthrough *is*. The explicit key exists so the intent reads in `config.hjson`, and so a **typo** in `areaTag` is not quietly reinterpreted as "carry this in transit".

The payload goes to a per-echo directory under `paths.ticTransit` and is queued from there. Everything downstream is untouched: forwarding queues `localInfo.newPath` whatever kind of path it is.

## Cleanup is less blocked than the issue assumed

The issue said expiry "has to be the thing that decides when a transit file may finally go", which is why this was parked behind #755. That turns out not to be true.

`inspectOutbound()` enumerates every reference across every node, so **"does anyone still owe this file?" has an exact answer**. A file a downlink has not collected is named by a live flow file line and stays; one nothing names goes. That is precisely what an age threshold would have been approximating, and worse — a link that is merely offline still has live references, so its files are already kept for exactly as long as it takes. No policy, no threshold, and no dependency on #755.

A `~` (sent) reference still counts. Those lines are history, the flow file drains on its own, and deleting underneath one buys nothing while making "why is this gone?" unanswerable.

The sweep runs at the end of an import pass rather than after each forward — the answer only changes when a downlink *collects* something, and a file just queued is by definition still referenced.

## Duplicate protection, which the file base was doing for us

#743 relies on the file-base collision rename: a second announcement of a name we hold is stored under a renamed copy, and `forwardTicToDownlinks()` then refuses to forward it, because announcing one name while shipping another leaves the downlink an orphan.

A transit file **must** be stored under exactly the name it was announced as — BinkP offers a file by its basename and htick pairs strictly by name (the reasoning in #759) — so there is no rename to fall back on. The two cases are separated instead:

- **Still queued for some downlink** → refused and archived. Overwriting would change the bytes under a reference that is already queued, and the downlink would receive the new file under the old announcement's CRC.
- **Referenced by nobody** → a leftover the sweep has not reached. Replaced.

## `Replaces`

Matched by name inside the echo's own transit directory rather than against file base metadata. Origin scoping is not lost by that: `tic_origin` comes off the TIC and an attacker sets it — as the existing `canForwardTic()` comment already notes — so the boundary that does the work is `uplinks`, applied well before this. The per-echo directory is what stops a pattern reaching across echoes, and more than one match is refused rather than guessed at.

## Startup diagnostics

A passthrough area with no `downlinks` accepts files and then sweeps them — it is not a leaf, which is what "no downlinks" means for a stored area. And one with no `uplinks` is worse off than a stored area in the same state: it refuses everything *and* keeps nothing for local users. Both are now said out loud, and the existing no-uplinks warning is passthrough-aware rather than duplicated.

## Verified against a live instance

Driven through real import passes on a throwaway ENiGMA with `fileBase.areas: {}` — no file areas at all:

```
=== TRANSIT ===   ~/transit/fsx_gen/COOLAPP.ZIP
=== FLOW ===      ~/transit/fsx_gen/COOLAPP.ZIP
                  ^~/ob/outbound/11b05667.tic
=== TIC ===       Origin 21:1/100        (the hatcher, not us)
                  From 21:1/151  To 21:1/200  Pw DOWNPASS
                  Path 21:1/100 ... / Path 21:1/151 ...
                  Seenby 21:1/100 / 21:1/151 / 21:1/200
```

Then: sweep keeps it while queued, keeps it when marked `~`, removes it once the flow file drains; a re-announce while queued is refused with the old bytes intact; `Replaces COOLAPP.*` dequeues the old payload and its TIC and queues the new pair.

## Tests

+23 (2887 → 2910), 0 failing.

`test/tic_passthrough.test.js` drives the **whole import waterfall**, not a stub of it — possible precisely because passthrough touches no database, so what is left is configuration and the filesystem. (`ftn_tic_import.test.js` has to stub the half below validation for exactly that reason.)

Covered: carrying a file with no file base area at all, the inbound draining with nothing rejected, the outgoing TIC's `Origin`/`Path`/`Seenby`, all four sweep cases including another echo's directory being left alone, both re-announce cases, both `Replaces` cases, and the publish check still refusing a non-uplink.

Mutation-checked: turning `isPassthroughArea` off fails 8; removing the duplicate refusal fails 1 (the one that names it); making the sweep ignore references fails 7.


NuSkooler:

| Pair | Result |
|---|---|
| #863 + #865 | clean |

NuSkooler:
## Review fixes — all seven confirmed findings closed

An independent review of this branch found seven defects, two of them serious enough that it should not have merged. All are fixed in `caa5d4c0`; each reproduces through the real import waterfall and each has a test that fails without the fix.

| # | Defect | Fix |
|---|---|---|
| 1 | **Remote DoS.** `globToRegExp` mapped `*` → `.*`; a run backtracks exponentially (6 stars 43ms, 8 stars 1.5s, 10 stars **36s**), synchronously, on a value from a peer's TIC | Linear scanner. 36s → 0.011ms, all 21 semantic cases preserved |
| 2 | **Silent file loss.** A busy `.bsy` made the forward fail; the import still "succeeded", the inbound copies were unlinked, and the sweep deleted the transit copy | A forward that attempted downlinks and queued none defers: transit copy removed, TIC held for retry under `holdMaxAgeMs` |
| 3 | **Upper-case `ticAreas` key.** Three lookups, only two lower-cased — stored, never forwarded, then swept | One shared case-insensitive resolver |
| 4 | **Same-name `Replaces` refused** — the weekly nodelist case, rejected every week while a downlink stayed offline | Overwrites in place, as the stored path does |
| 5 | **`Sha256` TIC forwarded with no `Crc`**, a required field, so downlinks rejected it | `validate()` always computes the CRC; still only *verifies* it when there is no `Sha256` |
| 6 | **The sweep trusted an incomplete answer** — `inspectOutbound` skips an unreadable flow file, so that node's files looked unwanted | It now reports skips, and the sweep refuses to run that pass |
| 7 | **A working config would have become a deleting one** — passthrough was inferred from an absent `areaTag`, but such an entry beside a `fileBase` area is a stored area that works today | `passthrough: true` required, and must be a literal `true` (hjson hands you `"false"` for an unquoted value) |

### One more, and it is not passthrough-specific

Fixing #4 exposed that `dequeueReplacedForDownlinks()` returns early for a file replaced **in place**, reasoning that the reference is still good. The reference is — the *TIC queued alongside it* is not, since it announces the superseded file's `Crc` and `Size` for bytes that are now the new file's. The downlink was offered two announcements for one payload and would reject the stale one. Now dequeued and re-queued on that path too, which fixes it for **stored areas as well** — a live defect in shipped #743.

### Verification

Full suite **2920 passing, 0 failing** (+33 over the 2887 baseline). Six mutations, one per fix, each caught by the test that names it.

Two of those mutations initially survived, which is the part worth recording:

- The ReDoS test exercised the matcher directly but not the call site. A linear matcher is only worth something if it is the thing actually called — the test now drives a hostile `Replaces` through a real import pass.
- The `Crc` test asserted a line matching `[0-9A-F]{8}` existed. With the digest never fed, `crc.finalize()` returns `00000000` — eight hex digits, and a lie. It now computes the expected CRC from the payload and compares values.

Also re-verified against a live instance: an upper-case `ticAreas` key forwards; a same-name weekly supersede leaves one TIC queued and nothing rejected; a hostile `Replaces` completes a whole import pass in ~1.1s including node startup; and a file delivered while the downlink held its `.bsy` stays in the inbound and is delivered on the next pass rather than vanishing.

PR #867, opened 2026-09-15
feat(tic): address each downlink from our AKA in its own network

### LINKED ISSUE #757: TIC: support a file area carried on more than one network
Follow-on from #743, recorded there as a declared v1 constraint.

A `ticAreas` entry names a single `network`, which supplies the local address that signs the outgoing `From`, `Path` and `Seenby`. A file echo carried on two networks — or a hub with several AKAs wanting to present the right one per downlink — cannot be expressed.

## Prior art

- htick selects the AKA per link (`downlink->ourAka`).
- Synchronet's tickit does AKA distance-matching against the link, with per-area and global overrides.

## Scope

- Per-downlink AKA selection rather than one address per area.
- Sensible default: closest AKA by zone/net, as tickit does.
- The loop guard must keep treating **all** our addresses as "us" across every network — #743 already does this via `Address.isAnyOf`, so that part is in place.
- `resolveNetworkNameForZone` in `core/bso_util.js` already infers a network from a downlink's zone and warns when more than one claims it; that is the natural place to extend.

Currently the startup diagnostics warn when an area's zone is claimed by more than one network, which is the symptom this would resolve properly.

NuSkooler:
Closes #757.

> **Merge note:** this touches `forwardTicToDownlinks()` / `queueTicForDownlinks()`, which #865 (hatching) also restructures. They will conflict; #866 (passthrough) does not touch either. Whichever lands second needs a rebase — the change is small in both directions, and the shapes are compatible: #865 extracts the half after the gates into `announceTicToDownlinks()`, and this makes the identity inside it per-downlink.

## The defect

A `ticAreas` entry named a single `network`, whose `localAddress` signed the outgoing `From`, `Path` and `Seenby` for every downlink.

Worse than the issue describes: when no `network` was named, `getNetworkNameForTicArea()` took one from the **first downlink's zone** and used it for all of them. So an echo fed to a Fidonet link and an fsxNet link announced both under whichever address happened to come first, and the link in the other network received a TIC from an address it does not know us by.

## The fix

The `From` line, our `Path` line and the outbound directory now follow **one per-downlink choice**, together, so they cannot disagree — a mailer dialling the right node from the wrong identity is exactly the failure worth avoiding here.

`selectLocalNetworkForAddress()` in `bso_util` does the choosing, by AKA distance:

| Distance | |
|---|---|
| 0 | same zone **and** same net — a neighbour in their own net |
| 1 | same zone — same network, different net |
| 2 | neither — a last resort, and warned about |

That is Synchronet tickit's rule, and what htick does per link with `downlink->ourAka`. Ties go to `defaultNetwork` then config order, matching `resolveNetworkNameForZone()` so the two cannot disagree about a zone more than one network claims.

`resolveNetworkNameForZone()` itself is **unchanged**. The issue suggested extending it, but it answers "which network owns this zone?" — the right question for a *directory*, since a zone maps to exactly one outbound subdirectory, and the wrong one for a `From` line, because it gives a single answer for a whole area. The two questions now have two functions.

## Overrides

Most specific first:

1. `nodes.<address>.tic.network` — this link, file echoes only
2. `nodes.<address>.network` — this link; the key NetMail routing already uses
3. `ticAreas.<tag>.network` — every downlink of this area
4. Closest AKA

`ticAreas.<tag>.network` sits **above** distance matching deliberately. It is the knob #743 shipped and the one an operator has already used to pin an area; letting closeness silently override it would change behaviour under a configuration that was explicit about what it wanted. A node-level setting is more specific and overrides it, which is how one link is carved out of a pinned area. A network name that is not configured falls back to the closest AKA rather than costing the downlink its file, and says so.

## Seenby

`Seenby` now names **every AKA in play in the echo**, not one.

It is the loop guard, and a peer matches it against the address *it* knows us by — tickit does so with literal string equality, which is why the existing code is already careful to write it 4D. A peer that knows us by our Fidonet address would not recognise an fsxNet-only `Seenby` and would forward the file straight back at us. All of our AKAs have genuinely seen the file, so listing them is both true and the safe direction to err.

Scoped to the AKAs actually in play rather than every address we own: an address on a network the echo is not carried on has nothing to do with the file and would propagate as noise into every downstream TIC. **A single-network system sees no change at all** — there is one AKA and it is already the one listed.

Downlink selection still treats all our addresses as us across every network, via the existing `Address.isAnyOf`.

## Also

The `Replaces` dequeue follows the same choice. A superseded file lives in whichever network directory it was filed under, so scrubbing the area's directory for every downlink would silently leave it queued for the rest.

Startup reports the AKA each downlink will be addressed from, and warns when no address of ours shares its zone — the answer is a property of the configuration, and an operator should be able to see it before a file goes out under the wrong name rather than after.

## Tests

+24 (2887 → 2911), 0 failing.

- `test/bso_util.test.js` — the distance rule: same-net, same-zone, per-network answers for three networks, a zone nothing shares, `defaultZone` beating `localAddress`, tie-breaking, an unparsable `localAddress`, and refusing to guess with nothing to go on.
- `test/ftn_tic_multi_network.test.js` — what lands in the spool for a two-network area: per-downlink `From`, `Path` and outbound directory (including `defaultNetwork` moving which one owns the bare `outbound/`), per-link passwords, both `Seenby` properties, all four override levels, the unconfigured-network fallback, the `Replaces` dequeue reaching both directories, and a single-network system behaving exactly as before.

Mutation-checked: collapsing the identity back to one network fails 6; reducing `Seenby` to a single AKA fails the one test that names it.


NuSkooler:

| Pair | Result |
|---|---|
| #863 + #865 | clean |

NuSkooler:
## Review fixes

Four defects found on review, closed in `d32f1ffd`. The first is the one that matters and it changes the shape of the PR.

### 1. Per-downlink *directory* selection broke the FTS-5005 lock

`BsoSpool` derives the canonical per-node `.bsy` path from the address alone (`_networkNameForAddr`, by zone) and is constructed with `networks` and `paths` and nothing else — it cannot see a node's `tic.network` or a `ticAreas` entry, so it could not agree with a per-downlink choice even in principle. A per-node lock cannot be derived from a per-area choice in any case: one node may be a downlink of several areas.

Filing a flow file anywhere else meant the tosser and a live BinkP session taking **different `.bsy` files**, so the exclusion #749 exists to provide stopped excluding anything — and `oputil bso prune --yes` would rewrite under a live session. An external binkd configured for the previous directories would also never look in the new one.

**Selection is now identity only** — `From`, our `Path` line, our `Seenby` entry. Nothing is lost by the split: an outbound session presents *every* one of our addresses in the handshake (`localAddresses`, `core/binkp/util.js`), so where a file sits has never had anything to do with the identity we announce it under.

Going the other way, the directory is now canonical **by the downlink's zone** rather than the area's network, with the rule shared with `BsoSpool` through `bso_util` so the writer and the mailer cannot drift — #719's lesson applied to the same class of bug. That also closes a **pre-existing** hole I had intended to file separately: an area with an explicit `network` already filed a downlink outside that network where the lock resolver would not look, and the inferred case (take the first downlink's zone, use it for all of them) did the same *by default*. The test that pins this asserts against `BsoSpool._bsyPath()` rather than a literal path, so the two cannot drift apart silently.

### 2. `nodes.<addr>.network` is no longer consulted

It is a pre-existing key read for NetMail routing. Treating it as a TIC setting changed the identity announced to a link under a configuration nobody edited. `tic.network` is the TIC-scoped one and stays.

### 3. A comment that was wrong, and a test that could not fail

The comment claimed this agrees with `resolveNetworkNameForZone()`. It doesn't — distance resolves on *net* before any tie-break is reached. And the test "proving" the tie-break used two networks **in the same net**, so every candidate was distance 0 and the tie-break necessarily ran: the assertion's premise could not have been violated. Both functions' behaviour is now stated correctly and both are tested, including the case where they disagree — which is exactly why only the zone resolver may decide a directory.

### 4. A 2D downlink was silently dropped

`103/999` with no zone is legal and handled deliberately elsewhere, but matches no network by zone. The comment claimed a fallback that existed only on the dequeue path. It now falls back to the area's address — in **one** place rather than two, because two fallbacks mean neither is testable.

### Verification

Full suite **2913 passing, 0 failing**. Six mutations; five caught:

| | |
|---|---|
| identity collapses to one address per area | 6 failures |
| directory follows identity again | 1 |
| `Seenby` back to a single AKA | 1 |
| node-level NetMail key repurposed | 1 |
| 2D downlink dropped | 1 |
| spool stops sharing the canonical rule | **0 — expected** |

The last survives correctly: that mutation swaps the shared helper for behaviourally identical code. Sharing guards against *future* drift, which no mutation test can detect.

Two of these initially survived and both were my tests passing for the wrong reason — the 2D fixture set an area `network`, so it returned on the explicit branch and never reached the fallback it claimed to test; and the duplicate fallback in `queueTicForDownlinks` masked the one being mutated. Both fixed.

PR #872, opened 2026-09-17
fix(menu): drain queued interrupts on arrival only

NuSkooler:
Based on #869, and a sibling of #870 rather than stacked under it — it touches only `menu_module.js` and `sysop_chat.js`, so it merges with the rest of the stack cleanly (verified: no conflicts, 2925 passing on a trial integration branch).

### The problem

The interrupt queue is drained at **four** sites: once on arrival in `initSequence()`, and three times on departure in `prevMenu()`, `nextMenu()` and `autoNextMenu()`.

The departure drains are the defect behind the corrupted WFC in #869, and the same one bit `sysop_chat`. They run before `menuStack` reaches `leave()`, so the outgoing menu's form is still attached to `'key press'` while `pausePrompt()` waits on a non-exclusive `once('key press')` — the key that dismisses an interrupt also fires that menu's action keys — and any refresh timer it owns keeps painting over the interrupt art.

#869 fixed this per module, with `prevMenu()` overrides on the WFC and `sysop_chat`. This is the general fix: **drain only on arrival**, where the destination owns the screen, nothing is attached and no timers are running.

### Why this is cheap rather than a 29-module audit

The arrival drain moves from `initSequence()` to `enter()`.

29 modules override `initSequence()` without draining, so a queued item was never shown while inside any of them. `enter()` is overridden by exactly four — `wfc`, `msg_list`, `file_area_list`, `msg_area_post_fse` — and **every one calls `super.enter()`**, so this reaches everything.

None of the four draws anything before that call. The only one doing work after it (`msg_list`, assigning config) now does so *before* `initSequence()` rather than racing it, which is an improvement.

So this is **strictly more coverage than before, not less**: the modules that never drained now do, on the way in. The "some menus would go quiet" worry was an artefact of the drain sitting somewhere 29 modules bypass.

### Opting out

A module that must not be interrupted on arrival overrides `displayQueuedInterruptions()` — the WFC routes instead of displaying (#870), and `sysop_chat` now defers, since draining there would paint over a conversation before it is drawn — or sets `interrupt: never`, which `displayQueuedInterruptions()` already short-circuits.

### One judgement call worth reviewing

Leave a menu, enter a module that defers, then disconnect, and those items are never shown; previously they would have been flushed on the way out. They are ephemeral session notifications, so I have taken that as acceptable — but it is a decision, and this is the place to disagree with it.

### Testing

2894 passing on this branch; 2925 merged with the rest of the stack.

Unit tests pin the contract: `prevMenu()`/`nextMenu()` hand the queue straight to the stack, `enter()` drains before the sequence starts, a module overriding `initSequence()` still gets its drain, and `interrupt: never` is still respected.

Live on a real board, a node message queued while inside `onelinerz`, `rumorz` and the bbs list still surfaces on every traversal, with nothing lost — verified both with and without this change, and again with the whole stack merged on top.

### Follow-up, not done here

`wfc.js` and `page_sysop.js` call `displayQueuedInterruptions()` as the first step of their own `initSequence()`. With the drain in `enter()` that is now a redundant second call — harmless, since the queue is already empty, but worth removing once this and #870 are both in.


NuSkooler:
Added the follow-up that was listed as out of scope, since it is the same logical change and this PR is still unreviewed.

`wfc.js` and `page_sysop.js` called `displayQueuedInterruptions()` as the first step of their own `initSequence()`. With the drain in `enter()` those only ever find an empty queue. Harmless, but it reads as though a module still has to do this itself — which is precisely the belief that made the departure drains hard to spot in the first place.

The merged tree now has exactly one drain site (`MenuModule.enter()`) plus the two deliberate opt-outs: `sysop_chat` defers, the WFC routes.

Re-verified merged with the rest of the stack — no conflicts, 2925 passing, and live on a real board: WFC routing and the message viewer, a queued message surviving traversal through `onelinerz`/`rumorz`/the bbs list, and a message arriving mid-chat still held and landing in the inbox afterwards.

One gap I could not stage live: an item queued *before* arriving at the WFC, to exercise the routed arrival drain. Every menu now drains on arrival, and this board's main menu is `interrupt: realtime`, so an item is always consumed before it can reach the dashboard. That path is covered by unit tests only.

